PR #28, opened 2024-07-04
chore: Adds a text component 

### LINKED ISSUE #17: Create shared Text Component
Create a shared text component that follows the design patterns of [CoMapeo Mobile.](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/sharedComponents/Text.tsx).

## To Do
- [x] Create `Text` component in `sharedComponents` folder
- [x] Accepts styles but has a default styling **to be confirmed with Aldo**
- [ ] Uses [styled()](https://mui.com/system/styled/) for style overrides

### Props
```ts
type TextProp = {
     // check if it is a MessageDescriptor. If it is a message descriptor translate it using [formatMessage](https://formatjs.io/docs/react-intl/api#formatmessage)
     children : React.ReactNode || MessageDescriptor //https://formatjs.io/docs/react-intl/components#message-descriptor
     style: React.CSSProperties
}

```

cimigree:
closes #17 
### Description
- Creates a shred Text component that accepts a prop kind, that returns either 'title' | 'body'  | 'subtitle'. If no kind if passed, it defaults to 'body'.
- accepts props for bold, italic, underline
- Can pass style to any of the text components but these will not override the bold, italic, underline

### Small fixes
- Adds no text transform to the button as specified in the design docs (https://www.notion.so/digidem/Typography-a1600603854445ad82b8d67d4067be1a)
- Adds black to colors

Screenshot 2024-07-04 at 5 22 02 PM

### DIFF
--- a/src/renderer/components/Text.tsx
+++ b/src/renderer/components/Text.tsx
@@ -10,17 +10,6 @@ const kindToVariant: { [k in Kind]: Variant } = {
   body: 'body1',
 } as const
 
-const getTextStyles = (
-  underline?: boolean,
-  bold?: boolean,
-  italic?: boolean,
-  style?: CSSProperties,
-) => ({
-  textDecoration: underline ? 'underline' : style?.textDecoration,
-  fontWeight: bold ? 'bold' : style?.fontWeight,
-  fontStyle: italic ? 'italic' : style?.fontStyle,
-})
-
 type BaseProps = PropsWithChildren<{
   align?: Extract<
     CSSProperties['textAlign'],
@@ -46,14 +35,14 @@ export function Text({
   underline,
   ...otherProps
 }: TextProps) {
-  const textStyles = getTextStyles(underline, bold, italic, style)
-
   return (
     <Typography
       variant={kindToVariant[kind]}
+      fontWeight={bold ? 'bold' : undefined}
+      fontStyle={italic ? 'italic' : undefined}
       style={{
         ...style,
-        ...textStyles,
+        textDecoration: underline ? 'underline' : undefined,
       }}
       {...otherProps}
     />
--- a/src/renderer/components/Text.tsx
+++ b/src/renderer/components/Text.tsx
@@ -1,4 +1,3 @@
-// Text.tsx
 import * as React from 'react'
 import { PropsWithChildren } from 'react'
 import Typography, { TypographyProps } from '@mui/material/Typography'
--- a/src/renderer/components/Text.tsx
+++ b/src/renderer/components/Text.tsx
@@ -1,0 +1,78 @@
+// Text.tsx
+import * as React from 'react'
+import { PropsWithChildren } from 'react'
+import Typography, { TypographyProps } from '@mui/material/Typography'
+
+type BaseProps = PropsWithChildren &
+  TypographyProps & { style?: React.CSSProperties }
+
+export function TitleText({ children, style, ...otherProps }: BaseProps) {
+  return (
+    <Typography variant="h1" style={style} {...otherProps}>
+      {children}
+    </Typography>
+  )
+}
+
+export function BodyText({ children, style, ...otherProps }: BaseProps) {
+  return (
+    <Typography variant="body1" style={style} {...otherProps}>
+      {children}
+    </Typography>
+  )
+}
+
+export function DescriptionText({ children, style, ...otherProps }: BaseProps) {
+  return (
+    <Typography variant="body2" style={style} {...otherProps}>
+      {children}
+    </Typography>
+  )
+}
+
+export function SubtitleText({ children, style, ...otherProps }: BaseProps) {
+  return (
+    <Typography variant="subtitle1" style={style} {...otherProps}>
+      {children}
+    </Typography>
+  )
+}
+
+type TextProps = BaseProps & {
+  kind?: 'title' | 'body' | 'description' | 'subtitle'
+}
+
+export function Text({ children, kind, style, ...otherProps }: TextProps) {
+  switch (kind) {
+    case 'title':
+      return (
+        <TitleText style={style} {...otherProps}>
+          {children}
+        </TitleText>
+      )
+    case 'body':
+      return (
+        <BodyText style={style} {...otherProps}>
+          {children}
+        </BodyText>
+      )
+    case 'description':
+      return (
+        <DescriptionText style={style} {...otherProps}>
+          {children}
+        </DescriptionText>
+      )
+    case 'subtitle':
+      return (
+        <SubtitleText style={style} {...otherProps}>
+          {children}
+        </SubtitleText>
+      )
+    default:
+      return (
+        <Typography style={style} {...otherProps}>
+          {children}
+        </Typography>
+      )
+  }
+}
--- a/src/renderer/components/Text.tsx
+++ b/src/renderer/components/Text.tsx
@@ -1,5 +1,25 @@
 import { CSSProperties, PropsWithChildren } from 'react'
+import { type Variant } from '@mui/material/styles/createTypography'
 import Typography from '@mui/material/Typography'
+
+type Kind = 'title' | 'subtitle' | 'body'
+
+const kindToVariant: { [k in Kind]: Variant } = {
+  title: 'h1',
+  subtitle: 'subtitle1',
+  body: 'body1',
+} as const
+
+const getTextStyles = (
+  underline?: boolean,
+  bold?: boolean,
+  italic?: boolean,
+  style?: CSSProperties,
+) => ({
+  textDecoration: underline ? 'underline' : style?.textDecoration,
+  fontWeight: bold ? 'bold' : style?.fontWeight,
+  fontStyle: italic ? 'italic' : style?.fontStyle,
+})
 
 type BaseProps = PropsWithChildren<{
   align?: Extract<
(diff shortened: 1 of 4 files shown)


src/renderer/components/Text.tsx
```diff
@@ -0,0 +1,77 @@
+import * as React from 'react'
+import { PropsWithChildren } from 'react'
+import Typography, { TypographyProps } from '@mui/material/Typography'
+
+type BaseProps = PropsWithChildren &
```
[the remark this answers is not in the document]
achou11: not sure if we should extend MUI's props here because then the exported components are more closely tied to MUI. would be selective and only specify the props that are most relevant for our use cases and specify them without referencing MUI (is possible). Example:

```tsx
type BaseProps = PropsWithChildren<{
  align?: Extract<
    CSSProperties['textAlign'],
    'inherit' | 'left' | 'center' | 'right' | 'justify'
  >
  bold?: boolean
  className?: string
  id?: string
  italic?: boolean
  style?: CSSProperties
  underlined?: boolean
}>

export function TitleText({
  align,
  bold,
  children,
  className,
  style,
}: BaseProps) {
  return (
    <Typography
      variant="h1"
      align={align}
      style={style}
      fontWeight={bold ? 'bold' : undefined}
      className={className}
    >
      {children}
    </Typography>
  )
}
```

not exactly sure what kind of props should be exposed for end usage. the ones listed in the example are probably sufficient for now, but could imagine wanting other things like data attributes or aria attributes in the future

src/renderer/components/Text.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <SubtitleText style={style} {...otherProps}>
+          {children}
+        </SubtitleText>
+      )
+    default:
+      return (
+        <Typography style={style} {...otherProps}>
+          {children}
+        </Typography>
+      )
```
achou11, line 75: wondering if this should throw an error instead, as it means that the end user is trying to use this component in a way that isn't represented by the `kind` prop.

src/renderer/pages/Home.tsx
achou11, line 1: My understanding is that the `Button` will handle the `Typography` internally, so I don't think there's a need to use the `Text` component within a `Button` in most cases (guess it lets you do a style override)

src/renderer/components/Text.tsx
```diff
@@ -0,0 +1,77 @@
...
+export function SubtitleText({ children, style, ...otherProps }: BaseProps) {
+  return (
+    <Typography variant="subtitle1" style={style} {...otherProps}>
+      {children}
+    </Typography>
+  )
+}
+
+type TextProps = BaseProps & {
+  kind?: 'title' | 'body' | 'description' | 'subtitle'
```
achou11, line 41: think `description` is kind of vague, especially given `body` (when do i know which to use?). maybe it can be omitted for now and reintroduced with some clearer name if needed?

achou11 (changes requested):

gmaclennan:
Optional of course, but there is the DRY option, although sometimes these can be too "clever" vs. just typing things out.

```ts
const mapKindToMUIVariant = {
  body: "body1",
  title: "h1",
  subtitle: "subtitle1",
} as const
  

export function Text({
   kind 
   bold,
   italic,
   style,
   underlined,
   ...passThroughProps
 }: TextProps) {
   const muiVariant = mapKindToMUIVariant[kind]
   if (!muiVariant) throw ...
   return (
     <Typography
       variant={muiVariant}
       style={{
         ...style,
         fontWeight: bold ? 'bold' : undefined,
         fontStyle: italic ? 'italic' : undefined,
         textDecoration: underlined ? 'underline' : undefined,
       }}
       {...passThroughProps}
     />
   )
```

src/renderer/components/Text.tsx
```diff
@@ -0,0 +1,127 @@
...
+}: BaseProps) {
+  return (
+    <Typography
+      variant="h1"
+      align={align}
+      style={{
+        ...style,
+        fontWeight: bold ? 'bold' : undefined,
+        fontStyle: italic ? 'italic' : undefined,
+        textDecoration: underlined ? 'underline' : undefined,
```
gmaclennan, line 34: This means that you if you pass the style prop `{ fontVariant: 'bold' }`, this will be overridden with `undefined` for the prop passed to `Typography`. Is this desired behaviour?

achou11:
> Optional of course, but there is the DRY option, although sometimes these can be too "clever" vs. just typing things out.

yeah upon further reflection, this may be fine for our needs. i had suggested the more repetitive approach because it had advantages that - in retrospect - may not be super relevant:

- can import specific text kind components directly, in the case where you don't need the "dynamism" of a singular text component (e.g. unlikely you'd want to switch some text between title and body styling)

- easier to implement different behavior if needed because of the separation in implementation. think it would be pretty rare that this would happen (why would i want different component behavior between e.g. title and body text?? 😄 )

src/renderer/components/Text.tsx
achou11: just to clarify, do you mean `fontWeight` in the example you provided? ([font variant](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant) is a separate but unrelated css property) I'm assuming that's what you meant, and thus the answer is no, we should definitely attempt to fall back to the `style` prop if provided e.g.

```tsx
fontWeight: bold ? 'bold' : style?.fontWeight,`
```

an additional thing that should be clarified is the precedence of props like `bold` vs `style`. not sure how these libraries usually behave, but what should happen when you have something like this?

```tsx
<Text bold style={{ fontWeight: 'normal' }} />
```

gmaclennan: yeah fontWeight is what I meant. Precedence is tricky! Also, does MUI actually support the style prop? I feel like we should keep the wrapper on MUI fairly light right now - my main question would be "how does MUI handle bold"?

gmaclennan:
> * can import specific text kind components directly, in the case where you don't need the "dynamism" of a singular text component (e.g. unlikely you'd want to switch some text between title and body styling)

I think this is just an API preference - `<Text variant="body"/>` vs `<BodyText />` - I see different component libraries choose one of the other. I don't mind. You can still do this DRY with `const BodyText = (props) => <Text kind="body" {...props} />` but repitition also fine!

> * easier to implement different behavior if needed because of the separation in implementation. think it would be pretty rare that this would happen (why would i want different component behavior between e.g. title and body text?? 😄 )

I don't think we should try to think through too many "possible scenarios" until we actually need them. Let's try and keep the surface area of the components we build to the minimum and add as needed, so if we do change MUi there is not much that we are depending on in the actual code.

src/renderer/components/Text.tsx
achou11: > does MUI actually support the style prop?

it supports a prop called [`sx`](https://mui.com/system/getting-started/the-sx-prop/) which is tied to style overrides via emotion's styled components. we opted to map our `style` prop to that in the button component (see https://github.com/digidem/comapeo-desktop/pull/27#discussion_r1664589542) with the intention of rarely using it

achou11: ah it also supports a `style` prop, which gets passed all the way down to the html element without any of the fanciness of that `sx` prop

achou11: > my main question would be "how does MUI handle bold"?

looks like both `style` and `sx` override any of the more direct props in MUI, so i guess maybe just stick with that behavior? (which seems to make enough sense in general)

cimigree: When you say "stick with that", you mean stick with using style to override? @achou11 ?

src/renderer/components/Text.tsx
```diff
@@ -0,0 +1,127 @@
...
+  children,
+  className,
+  italic,
+  style,
+  underlined,
+}: BaseProps) {
+  return (
+    <Typography
+      variant="subtitle1"
+      style={{
```
achou11, line 81: applies to all variants, but seeing our custom `style` prop being mapped to MUI's `style` prop makes me wonder if we should update our `Button`  to do the same thing. Currently it exposes a `sx` prop that maps to MUI's `sx` prop, which has some MUI-specific magic going on (which we probably don't really need).

Mostly want to be consistent across our components when it comes to style overrides. I think just the exposing `style: CSSProperties` like this text component is doing feels safer to me. any thoughts?

achou11 (changes requested):
discussed some of this elsewhere but for the sake of transparency: think Gregor's suggestion of a single component implementation that maps the kind to variant is more reasonable to work with. i changed my mind as I found out that the tediousness of implementing each individual components was not worth the small benefits the approach provides 😅

src/renderer/components/Text.tsx
achou11: yep! hoping this [comment](https://github.com/digidem/comapeo-desktop/pull/28#discussion_r1674436563) clarifies a bit but yeah I think a reasonable rule of thumb for any components that expose a `style` prop would be that it overrides any props that may be affected by that `style` prop (in our case, `bold`, `italic`, and `underlined`)

cimigree:
OK I think this is what you want?

src/renderer/components/Text.tsx
```diff
@@ -0,0 +1,50 @@
...
+  ...otherProps
+}: TextProps) {
+  return (
+    <Typography
+      variant={kindToVariant[kind]}
+      fontWeight={bold ? 'bold' : undefined}
+      fontStyle={italic ? 'italic' : undefined}
+      style={{
+        ...style,
+        textDecoration: underline ? 'underline' : undefined,
```
achou11, line 45: should account for `style` potentially specifying text decoration and give precedence to it:

```suggestion
      style={{
        textDecoration: underline ? 'underline' : undefined,
        ...style,
      }}
```

achou11 (approved):
one small area that should be addressed but otherwise lgtm 👍

achou11 (approved):

Issue #42, opened 2024-11-21
Onboarding Screens


ErikSin:
Create the onboarding/intro screen.

## Screens to build:
- [ ] Welcome
- [ ] Name Device
- [ ] Welcome: Data & Privacy UI
- [ ] Join or Create Project

### Images:
![Image](https://github.com/user-attachments/assets/da1d2ee9-456a-4dc8-adfc-2b4ebaafc2cd)

![Image](https://github.com/user-attachments/assets/7132ee65-6e04-4168-bdfb-1e822719fedf)

![Image](https://github.com/user-attachments/assets/2e79fe5a-bdef-4d14-9442-baced982f6f1)

![Image](https://github.com/user-attachments/assets/9b11f4d5-5bf3-429c-a7bc-f52846bc55a8)

### Figma:
https://www.figma.com/design/UIA00RSP8Sz24V0vXvQHeL/(Desktop)-CoMapeo?node-id=0-1&node-type=canvas&t=5TAgvrETKEGDVuVy-0

PR #49, opened 2024-12-03
feat:  device naming screen

cimigree:
related to #42 

### Description

- Adds the device naming screen.
- Adds the first query to set the device name
- Adds a disabled prop for the button component so it can be disabled during the is pending part of the data submission
- Adds the next screen (create join project screen) to make sure that the device name comes through
- For the next PR, I will make a reusable component to represent the Container and content box.

### Screenshots

**Blank Screen**

Screenshot 2024-12-04 at 11 03 59 AM

**Error State if you try to submit the name when there is no name**

Screenshot 2024-12-04 at 11 04 15 AM

**Character Count**

Screenshot 2024-12-04 at 11 04 44 AM


EvanHahn (approved):
LGTM other than some nitpicks and something that tripped my Unicode alarm.

src/renderer/src/images/device.png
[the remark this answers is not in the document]
cimigree: Good eye. I fixed the proportions to match Figma.

src/renderer/src/queries/deviceInfo.ts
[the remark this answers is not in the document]
cimigree: It does copy mobile app, so all good.

src/renderer/src/routes/Onboarding/CreateJoinProjectScreen.tsx
```diff
@@ -0,0 +1,43 @@
...
+})
+
+function CreateJoinProjectScreenComponent() {
+	const { data } = useDeviceInfo()
+
+	return (
+		<Container>
+			<OnboardingTopMenu currentStep={3} />
+			<ContentBox>
+				<Text>Device name is {data?.name}</Text>
```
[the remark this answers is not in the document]
cimigree: This is just temporary code I included to check if the device name came through. It is not actually used here or on what will be here.

src/renderer/src/routes/Onboarding/DeviceNamingScreen.tsx
```diff
@@ -0,0 +1,233 @@
...
+import { defineMessages, useIntl } from 'react-intl'
+
+import { BLACK, BLUE_GREY, DARK_COMAPEO_BLUE, RED, WHITE } from '../../colors'
+import { Button } from '../../components/Button'
+import { OnboardingTopMenu } from '../../components/OnboardingTopMenu'
+import { Text } from '../../components/Text'
+import DeviceImage from '../../images/device.png'
+import { useEditDeviceInfo } from '../../queries/deviceInfo'
+
+const DEVICE_NAME_MAX_LENGTH = 60
```
EvanHahn, line 14: Nit: Maybe change this to `DEVICE_NAME_MAX_LENGTH_GRAPHEMES`?

src/renderer/src/routes/Onboarding/DeviceNamingScreen.tsx
```diff
@@ -0,0 +1,233 @@
...
+export function DeviceNamingScreenComponent() {
+	const navigate = useNavigate()
+	const { formatMessage } = useIntl()
+	const [deviceName, setDeviceName] = useState('')
+	const [error, setError] = useState(false)
+	const [errorMessage, setErrorMessage] = useState('')
+	const setDeviceNameMutation = useEditDeviceInfo()
+
+	function getGraphemeSegments(text: string): Array<string> {
+		if (typeof Intl !== 'undefined' && 'Segmenter' in Intl) {
```
EvanHahn, line 120: Because we control the environment, maybe we don't need this at all? I think `Intl.Segmenter` should always be defined.

src/renderer/src/routes/Onboarding/DeviceNamingScreen.tsx
```diff
@@ -0,0 +1,233 @@
...
+			const segmenter = new Intl.Segmenter(undefined, {
+				granularity: 'grapheme',
+			})
+			const segments = [...segmenter.segment(text)].map((s) => s.segment)
+			return segments
```
EvanHahn, line 125: We only ever care about the number of graphemes. Rather than create an array just to get its length, could we change this function to `getGraphemeCount` and do something like this?

```ts
function countGraphemes(text: string): number {
  const segmenter = new Intl.Segmenter(undefined, {
    granularity: 'grapheme',
  })
  let result = 0
  for (const _ of segmenter.segment(text)) result++
  return result
}
```

src/renderer/src/routes/Onboarding/DeviceNamingScreen.tsx
```diff
@@ -0,0 +1,233 @@
...
+		const byteLength = getUtf8ByteLength(value)
+		let error = false
+
+		if (
+			graphemeCount > DEVICE_NAME_MAX_LENGTH ||
+			byteLength > DEVICE_NAME_MAX_BYTES
+		) {
+			error = true
+		} else {
+			if (value.trim().length === 0) {
```
EvanHahn, line 148: I wonder if we should _always_ `trim`?

EvanHahn (approved):

PR #53, opened 2024-12-09
Feat: onboarding create join project

cimigree:
related to  #42 
### Description

- Moves some files around
- Adds bigger icons
- Adds components for the repeated background for the Onboarding Screens
- Updates existing screens to use the new OnboardingScreenLayout component
- Updates the Onboarding Top Menu so a user cannot bypass the necessary steps
- Updates the first Onboarding screen to be more responsive and styled better
- Adds the Create project Screen
- Adds the Join project screen

### Notes

- Many of the changed files are just moving files into folders and updating the imports and also updating the size of some icons
- I added the Toggle for importing a config but I wasn't sure if that was actually part of the design or if that is done differently on desktop so I didn't add any logic for it
- I was not sure when the Join Project screen should show up... If the logic was the same as on mobile (Like what is the idea? Someone will get to the Create or Join Project screen and someone else will be right there and then the invite will be sent?) The discovery will have to be there, right?
- I was thinking we could add the logic later and just start with the UI/ screens for now

so you can get to the screen to see it by changing the navigation in the CreateJoinProjectScreen
from:
`onClick={() => navigate({ to: '/Onboarding/CreateProjectScreen' })}`
to:
`onClick={() => navigate({ to: '/Onboarding/JoinProjectScreen' })}`


ErikSin:
I think we can simplify the onboarding screen and make it more testable and maintainable by exposing the children at the top level and avoiding such deep nesting.

I would suggest:
```tsx
// this screen now controls whether the onBack can be pressed, and it does not have to be determined in OnboardingTopMenu

// Also the screen content can be fully determined by the parent. No need to pass props to the OnboardingScreenLayout and conditionally render things based on what screen it is
const SomeOnboardingScreen = () => {

   function onBackPress(){
         navigate(to next screen)
    }
    return(
        <OnboardingScreenLayout
            topMenu={<OnboardingTopMenu onBackPress={ionBackPress) currentStep=}/>}
        >
          {...screen content here}
        </OnboardingScreenLayout>
    )
}

const OnboardingScreenLayout = ({OnboardingTopMenu: ReactNode, children:ReactNode}) => {

    return(
        <Container>
			<OnboardingTopMenu/>
			<ContentWrapper>
				<ContentBox>
                    {children}
                </ContentBox>
            </ContentWrapper>
        </Container>
    )
}

const OnboardingTopMenu = ({onBackPress?, currentStep}) => {

    return (
        <MenuContainer>
            <GoBackButton disabled={!onBackPress} onPress={onBackPress}>
            <Steps>
              {[1, 2, 3].map(step => <Step active={currentStep === step}/>)}
            </Steps>
        </MenuContainer>
    )
}

```

This suggested architecture does make the component slightly less reusable as It is simply a layout wrapper and the parent determines the content completely.

But any shared styling can still be shared, just pass it around to the children not encapsulated in the `OnboardingScreenLayout`

The first major thing that this does is it allows the OnboardingTopMenu to be completely agnostic and not reliant on the nav state. In the future if we change the steps, we do not need to refactor the OnboardingTopMenu to account for the changes in navigation. Since the parent just controls this we can simply add or take away a screen with no refactoring.

Secondly we can test the screens in isolation. The `OnboardingScreenLayout` is solely for stying, so there is no logic to be tested. Previously if we wanted to test the < input /> in the Device name screen, and how it interacted with the button, we would have to test the DeviceNamingSreen, and then test the button rendered inside the OnboardingScreenLayout

src/renderer/src/components/Onboarding/OnboardingTopMenu.tsx
```diff
@@ -82,37 +81,53 @@ export function OnboardingTopMenu({ currentStep }: OnboardingTopMenuProps) {
...
-							<Text kind="body" bold={currentStep === step}>
-								{formatMessage(m.step, { number: step })}
-							</Text>
-						</Step>
-						{step < 3 && <Divider />}
-					</StepsContainer>
-				))}
+				{[1, 2, 3].map((step) => {
+					const isDisabled =
+						step > currentStep || (currentStep === 3 && step < currentStep)
```
ErikSin, line 110: as mentioned in slack. lets just make none of these clickable as this logic is a bit confusing

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
```diff
@@ -0,0 +1,283 @@
...
+		if (
+			graphemeCount > PROJECT_NAME_MAX_LENGTH_GRAPHEMES ||
+			byteLength > PROJECT_NAME_MAX_BYTES ||
+			value.trim().length === 0
+		) {
+			localError = true
+		} else {
+			setProjectName(value)
+		}
+		setError(localError)
```
ErikSin, line 139: can we write a test for this?

we could write a pure function that does not deal with react state

Something like:

```ts
const handleChange = (event: ChangeEvent<HTMLInputElement>) =>{
     const value = event.target.value
     const localError:boolean= checkForError(value)
     if (!localError) setProjectName(value)
     if (localError !== error) setError(localError)
}
```

then we could write a test for `checkForError` which is not reliant on react state, and can be a pure function

ErikSin (changes requested):

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
cimigree: I think I did this. I tried to anyway.

cimigree:
There is still logic that needs to be added, but I think the component and screen structure might be closer to what you had in mind. I didn't make shareable components for the buttons or anything. Hope that is ok, but let me know if I should do that too.

src/renderer/src/routes/Onboarding/PrivacyPolicyScreen.tsx
```diff
@@ -21,18 +21,22 @@ const ContentBox = styled('div')({
...
+	const navigate = useNavigate()
 	return (
 		<Container>
-			<OnboardingTopMenu currentStep={1} />
+			<OnboardingTopMenu
+				currentStep={1}
+				onBackPress={() => navigate({ to: '/Onboarding/DataPrivacy' })}
+			/>
 			<ContentBox>
 				<PrivacyPolicy />
```
ErikSin, line 41: Why is the privacy policy a nested component. It basically a container inside a container. it just seems like an unnecessary nesting

src/renderer/src/routes/Onboarding/DeviceNamingScreen.tsx
```diff
@@ -51,110 +54,62 @@ export const Route = createFileRoute('/Onboarding/DeviceNamingScreen')({
...
-				error = true
-			}
-			setDeviceName(value)
+		const localError: boolean = checkForError(
+			value.trim(),
+			DEVICE_NAME_MAX_LENGTH_GRAPHEMES,
+		)
+		setDeviceName(value)
+		if (localError !== inputError) {
+			setInputError(localError)
```
ErikSin, line 104: if there is an error, do we want to set the name? Or should we just set the device name when no error

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
```diff
@@ -0,0 +1,282 @@
...
+							padding: '12px 0',
+							cursor: 'pointer',
+						}}
+						onClick={() => setAdvancedSettingOpen(!advancedSettingOpen)}
+					>
+						<Text bold style={{ fontSize: '1.125rem' }}>
+							{formatMessage(m.advancedProjectSettings)}
+						</Text>
+						{advancedSettingOpen ? <ChevronDown /> : <ChevronUp />}
+					</div>
```
ErikSin, line 229: I think we should us an [accordian here](https://mui.com/material-ui/react-accordion/). Accordian would give the animation, and doesn't require us to use css to hide

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
```diff
@@ -0,0 +1,282 @@
...
+		setConfigFileName('myProjectConfig.comapeocat')
+	}
+
+	const topMenu = (
+		<OnboardingTopMenu
+			currentStep={3}
+			onBackPress={() =>
+				navigate({ to: '/Onboarding/CreateJoinProjectScreen' })
+			}
+		/>
```
ErikSin, line 155: we should also disable the back button when `setProjectNameMutation.isPending`

src/renderer/src/components/Onboarding/onboardingLogic.ts
ErikSin, line 1: can you write test for these functions? Vitest was just integrated into main, so you can rebase and just start wrtiting test. We are going to colocate are test files (which I will document). So if you can create a `onboardingLogic.test.ts` file and test these function there

ErikSin (changes requested):
Last things, can yo write tests for your function. Vitest should now be setup!

src/renderer/src/routes/Onboarding/DeviceNamingScreen.tsx
cimigree: I am doing that because when there is a blank there (someone erases all of the letters), there is an error **and** we need to set the device name to a blank. If I don't set the device name in that case a user can never erase that last letter.

src/renderer/src/routes/Onboarding/PrivacyPolicyScreen.tsx
cimigree: I did that in the same way I did it on mobile because the Privacy Policy is used in a separate place within the settings and is almost exactly the same so that way it can be reused as a component there.

src/renderer/src/components/Onboarding/onboardingLogic.ts
cimigree: Done! That was fun! Much more fun than E2E tests!

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
cimigree: Sure. Done.

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
cimigree: Sounds good. Done.

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
```diff
@@ -0,0 +1,283 @@
...
+					style={{
+						width: '100%',
+						maxWidth: 400,
+						margin: '0 auto',
+						textAlign: 'center',
+					}}
+				>
+					<Accordion
+						expanded={advancedSettingOpen}
+						onChange={(_, expanded) => setAdvancedSettingOpen(expanded)}
```
ErikSin, line 219: One small thing. Lets make this [uncontrolled component](https://react.dev/learn/sharing-state-between-components). In otherwords, dont pass in the `expanded` and `onChange` props, and let it manage its own state. For maintainability, the less state the parent has to control the better.

ErikSin (approved):
Great job and nice tests! Just one small comment about uncontrolled vs controlled state. But otherwise good to merge

PR #54, opened 2024-12-11
chore: integrate testing

ErikSin:
Sets up vitetest and React Testing Library


package.json
```diff
@@ -44,7 +44,8 @@
 		"types:preload": "tsc -p src/preload/tsconfig.json",
 		"types:renderer": "tsc -p src/renderer/tsconfig.json",
 		"types": "npm-run-all --print-label --parallel --continue-on-error types:*",
-		"test": "npm-run-all --print-label --parallel lint types",
+		"test": "npm-run-all --print-label --parallel lint types test:vitest",
+		"test:vitest": "vitest --config ./src/renderer/vite.config.js",
```
achou11, line 48: the `test:vitest` script here will run vitest in watch mode, which I don't think is not desirable if being used as part of the `test` script (especially in the context of CI). I suggest:

- creating two scripts for running vitest: one for watching and the other for just running tests
- update the `test` script to reference the script that only runs the tests

```suggestion
		"test": "npm-run-all --print-label --parallel lint types vitest:run",
		"vitest:run": "vitest run --config ./src/renderer/vite.config.js",
		"vitest:watch": "vitest watch --config ./src/renderer/vite.config.js",		
```

Also updating the development docs to highlight the difference might be helpful

src/renderer/src/routes/Onboarding/DeviceNamingScreen.tsx
```diff
@@ -227,3 +223,7 @@ export function DeviceNamingScreenComponent() {
 		</Container>
 	)
 }
+
+export function getUtf8ByteLength(text: string): number {
+	return new TextEncoder().encode(text).length
+}
```
achou11, line 229: maybe outside the scope of this PR but would eventually move this to something like`renderer/lib/text.ts`

src/renderer/src/routes/(MapTabs)/_Map.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+vi.mock('@tanstack/react-router', () => ({
+	useNavigate: vi.fn(() => {
+		return { navigate: vi.fn() }
+	}),
+	createFileRoute: vi.fn(() => {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		return (options: any) => ({ component: options.component }) // Mocked implementation
+	}),
+	Outlet: () => <div>Mocked Outlet</div>,
+}))
```
achou11, line 15: just for my understanding, this has to be done for each test that relies on the router in some way?

achou11 (changes requested):
One suggestion that I think is worth implementing before merging but otherwise changes seem reasonable

src/renderer/src/routes/(MapTabs)/_Map.test.tsx
ErikSin: it depends. I think short answer yes. long answer, if we want to test actual navigation actions, we need to stub out the router instead of just stubbing out the function inside the router. 

I think more work needs to go into creating our mocks and stubs, but this just allows it run (its was basically a smoke screen test for jsdom)

package.json
ErikSin: https://github.com/digidem/comapeo-desktop/pull/54/commits/abd6bbaa9f036e8ae44f2c1c3595c2c3cf533828
https://github.com/digidem/comapeo-desktop/pull/54/commits/58760ff85e510cb4b37e970429ddf572818f258f

achou11 (approved):

PR #55, opened 2024-12-11
feat: import config

cimigree:
To be considered after https://github.com/digidem/comapeo-desktop/pull/53 is adopted and merged.
Adds the ability when creating a project to import a config file.
The easiest and most straightforward way to do this was through using html <import>


src/renderer/src/hooks/useConfigFileImporter.ts
```diff
@@ -0,0 +1,30 @@
+import { useState } from 'react'
+
+export function useConfigFileImporter() {
```
achou11, line 3: with some changes to how the file opening mechanics work (as detailed by my other comment), I can see this being simplified to a light wrapper for the "runtime" in the form of a react-query mutation. something along the lines of:

```ts
export function useSelectConfigFile() {
  return useMutation({
    mutationFn: async () => {
      // selectFile yet to be implemented
      return window.runtime.selectFile({
        extensionFilters: ['comapeocat']
      })
    }
  })
}
```

src/renderer/src/hooks/mutations/projects.ts
```diff
@@ -9,13 +13,22 @@ export function useCreateProject() {
...
+			}
 		},
-		onSuccess: () => {
+		onSuccess: (projectId: string) => {
 			queryClient.invalidateQueries({
 				queryKey: getProjectsQueryKey(),
 			})
+			queryClient.invalidateQueries({
+				queryKey: getProjectSettingsQueryKey({ projectId: projectId }),
+			})
```
achou11, line 31: this should not be needed because the invalidation call above this (using `getProjectsQueryKey()`) will cascade queries for a specific project as well

```suggestion
			queryClient.invalidateQueries({
				queryKey: getProjectSettingsQueryKey({ projectId: projectId }),
			})
```

achou11 (changes requested):
Think this PR is blocked on some changes we need to make at the Electron level - I'm going to spend a small amount of time putting up a separate PR to help this!

src/renderer/src/hooks/useConfigFileImporter.ts
```diff
@@ -0,0 +1,30 @@
...
+	const [fileName, setFileName] = useState<string | null>(null)
+	const [error, setError] = useState<string | null>(null)
+
+	function handleFileSelect(event: React.ChangeEvent<HTMLInputElement>) {
```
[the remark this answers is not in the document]
cimigree: @achou11 Sure that would be great if you can! If not, I can give it a go. Just let me know. Thanks!

src/renderer/src/hooks/mutations/projects.ts
cimigree: Ok done.

src/renderer/src/hooks/useConfigFileImporter.ts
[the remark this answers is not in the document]
cimigree: I used your new hook, removed the html input, renamed some things in the createprojectscreen, and adjusted the test.

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
```diff
@@ -243,41 +256,32 @@ function CreateProjectScreenComponent() {
...
-					disabled={setProjectNameMutation.isPending}
-				>
-					{setProjectNameMutation.isPending
-						? formatMessage(m.saving)
-						: formatMessage(m.addName)}
-				</Button>
-			</div>
+				{createProjectMutation.isPending
+					? formatMessage(m.saving)
+					: formatMessage(m.createProject)}
```
achou11, line 283:
```suggestion
				{formatMessage(
					createProjectMutation.isPending ? m.saving : m.createProject,
				)}
```

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
```diff
@@ -243,41 +256,32 @@ function CreateProjectScreenComponent() {
...
-								onClick={importConfigFile}
+								onClick={handleImportConfig}
 							>
 								{formatMessage(m.importConfig)}
 							</Button>
-							{configFileName && (
+							{configPath && (
 								<Text style={{ textAlign: 'center', marginTop: 12 }}>
-									{configFileName}
+									{configPath}
```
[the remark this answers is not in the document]
achou11: Might have existed prior to this PR, but there are some minor but noticeable noticeable layout issues when the text is sufficiently long (or the window sufficiently narrow).

image

image


potentially mitigated by only displaying the file name but the issues would exist regardless of that too (in more extreme cases)

src/renderer/src/hooks/useConfigFileImporter.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+		})
+
+		let selectedPath: string | undefined
+
+		await act(async () => {
+			await result.current.mutateAsync(undefined, {
+				onSuccess: (val) => {
+					selectedPath = val
+				},
+			})
```
achou11, line 39: Nit but specifying `onSuccess` is not necessary in this case when using `mutateAsync`:
```suggestion
			const val = await result.current.mutateAsync(undefined)
			selectedPath = val
```

src/renderer/src/hooks/useConfigFileImporter.test.tsx
```diff
@@ -0,0 +1,70 @@
...
```
achou11, line 65: similar comment about not needing to specify `onSuccess`

```suggestion
			const val = await result.current.mutateAsync(undefined)
			selectedPath = val
```

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
```diff
@@ -107,52 +108,64 @@ const HorizontalLine = styled('div')({
...
+		createProjectMutation.mutate(
+			{ name: projectName.trim(), configPath: configPath ?? undefined },
+			{
+				onSuccess: (projectId) => {
+					setActiveProjectId(projectId)
+					navigate({ to: '/tab1' })
+				},
+				onError: (error) => {
+					console.error('Error saving project:', error)
+					setErrorMessage(formatMessage(m.errorSavingProjectName))
```
achou11, line 162: saving a project can also fail if the config file is invalid e.g. you selected a file with the correct extension but it's not actually the correct format.

Quick suggestion is to update the surfaced error text as something that's more generic and not specifically about the name.

achou11 (changes requested):
couple of minor code suggestions and then also some UX-related feedback

src/renderer/src/routes/Onboarding/CreateProjectScreen.tsx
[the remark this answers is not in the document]
achou11: i think it would make sense to update `runtime.selectFile` to return both the file's path and name. can quickly put up a separate PR for that!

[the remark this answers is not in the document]
achou11: see https://github.com/digidem/comapeo-desktop/pull/81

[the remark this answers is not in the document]
cimigree: @achou11 Ok used it. Thanks!

src/renderer/src/hooks/useConfigFileImporter.test.tsx
```diff
@@ -0,0 +1,59 @@
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { act, renderHook } from '@testing-library/react'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { useSelectProjectConfigFile } from './mutations/file-system'
+
+describe('useSelectProjectConfigFile', () => {
```
achou11, line 7: non-blocking but would be good to add a test here for when the payload from `window.runtime.selectFile()` is invalid e.g. the `path` or `name` are not strings

src/renderer/src/hooks/useConfigFileImporter.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+		)
+
+		const { result } = renderHook(() => useSelectProjectConfigFile(), {
+			wrapper,
+		})
+
+		await act(async () => {
+			const val = await result.current.mutateAsync(undefined)
+			expect(val).toBeUndefined()
+		})
```
achou11, line 57: more of a question for my own education: is using `act` here necessary? the test would probably work the same without wrapping it, although maybe it's more idiomatic to use it because you'd do something similar to simulate [dispatching an event](https://react.dev/reference/react/act#dispatching-events-in-tests) 🤔

achou11 (approved):

src/renderer/src/hooks/useConfigFileImporter.test.tsx
cimigree: Short answer is that I got this error:
`Warning: An update to TestComponent inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */
`

I guess it is because the mutation triggers internal React state updates, and act ensures those updates are resolved before we do our assertions.

src/renderer/src/hooks/useConfigFileImporter.test.tsx
cimigree: Just FYI, to get it to "throw" the error, I have to mock the selectFile as a rejected promise... because otherwise I would actually have to spin up Electron or the preload script with an integration test or an E2E test. Hope that is ok. Is this useful? Not sure...

achou11: >  I have to mock the selectFile as a rejected promise

yeah this is what I figured you would do for now. think that's fine 👍

Issue #44, opened 2024-11-21
Add Tabs on main map screen


ErikSin:
Add [Tabs](https://mui.com/material-ui/react-tabs/#vertical-tabs) and integrate with tanstack router as the main navigation pages on the map screen. Reference these [unfinished figma pages.](https://www.figma.com/design/UIA00RSP8Sz24V0vXvQHeL/(Desktop)-CoMapeo?node-id=113-1307&t=5TAgvrETKEGDVuVy-4). the content of the tabs, and which tabs will exist has not been decided yet, so this ticket is just to create the scaffolding.

To Do:
- [ ] Create Vertical tabs on the map page
- [ ] Create Tab screen to render ontop of the mapview
- [ ] Tabs bar uses icons and text

Screenshots:
![Image](https://github.com/user-attachments/assets/02b19e06-d849-41c4-ac5f-4be6dca9e1f2)

Figma:
https://www.figma.com/design/UIA00RSP8Sz24V0vXvQHeL/(Desktop)-CoMapeo?node-id=113-1185&node-type=frame&t=Sxv9RapuDPmko5vI-0

ErikSin:
closed by #57

PR #57, opened 2024-12-11
chore: style navigation tabs

ErikSin:
Styled the navigation tabs to prep for the map component.

image


src/renderer/src/routes/(MapTabs)/_Map.tsx
```diff
@@ -1,34 +1,99 @@
...
+						value={'/tab2'}
+					/>
+					<MapTabStyled
+						icon={<InfoOutlinedIcon />}
+						label={
+							<Text style={{ fontSize: 10 }} kind="title">
+								{formatMessage(m.about)}
+							</Text>
+						}
+						value={'/tab2'}
```
cimigree, line 76: Having both settings and about map to tab2 is very confusing. Is this just temporary?

src/renderer/src/routes/(MapTabs)/_Map.tsx
```diff
@@ -1,34 +1,99 @@
...
+					/>
```
cimigree, line 77: If you add an sx here with paddingBottom the about icon won't be so close to the bottom of the screen...

src/renderer/src/routes/(MapTabs)/_Map.tab1.tsx
```diff
@@ -4,10 +4,10 @@ import { createFileRoute } from '@tanstack/react-router'
 import { Text } from '../../components/Text'
 
 export const Route = createFileRoute('/(MapTabs)/_Map/tab1')({
-	component: RouteComponent,
+	component: Observations,
```
cimigree, line 7: this file name is very confusing. Can we name it Map.Observations.tsx or Map.tabobservations.tsx or something?

src/renderer/src/routes/(MapTabs)/_Map.tab2.tsx
```diff
@@ -4,10 +4,10 @@ import { createFileRoute } from '@tanstack/react-router'
 import { Text } from '../../components/Text'
 
 export const Route = createFileRoute('/(MapTabs)/_Map/tab2')({
-	component: RouteComponent,
+	component: Settings,
```
cimigree, line 7: Same confusion about file naming.

src/renderer/src/routes/(MapTabs)/_Map.tsx
```diff
@@ -1,34 +1,99 @@
...
+					onChange={(_, value) => navigate({ to: value as MapTabRoute })}
+					orientation="vertical"
+				>
+					{/* PostAddIcon is a placeholder icon */}
+					<MapTabStyled
+						data-testid="tab-observation"
+						icon={<PostAddIcon />}
+						value={'/tab1'}
+					/>
+					<Box flexGrow={1} />
```
cimigree, line 58: This is kind of unfortunate that you had to do this to put the tabs at the bottom.

src/renderer/src/routes/(MapTabs)/_Map.tsx
```diff
@@ -1,34 +1,99 @@
...
+						<Outlet />
+					</Suspense>
+				</Box>
 			</Paper>
-			<div>map component here</div>
-			<div>parent map component render count: {renderCount.current}</div>
-		</div>
+			<Suspense fallback={<CircularProgress />}>
+				<div>map component here</div>
+			</Suspense>
```
cimigree, line 95: This component seems very complex. It handles routing, tab rendering, Suspense and layout. Could there be a separate function or hook to handle the logic of determining which tab is selected?
It would be nice if MapLayout could just focus on Layout

src/renderer/src/routes/(MapTabs)/_Map.tsx
```diff
@@ -1,34 +1,99 @@
...
+
+				<Box
+					sx={{
+						width: 300,
+						borderLeftColor: VERY_LIGHT_GREY,
+						borderLeftWidth: '1px',
+						borderLeftStyle: 'solid',
+					}}
+				>
+					<Suspense fallback={<CircularProgress />}>
```
cimigree, line 88: Why does Suspense need to be here twice? Are we always going to need a Suspense state at every level?

cimigree (changes requested):
Is this related to this issue? [chore-style-tabs](https://github.com/digidem/comapeo-desktop/issues/44)

I am getting tons of errors / warnings both in the terminal when running tests and the developer console (especially when clicking on the tabs) These relate to invalid props on Dom elements (fullWidth, indicator, selectionFollowsFocus, and textColor) that we are not using explicitly so not sure what is going on here. Did you look into it? It seems like it is something we should try to fix. If for no other reason than that any [other] errors we need to address will be invisible among all the other noise.

The other warning I get repeatedly is:
chunk-GYPSVNDI.js?v=08559733:674 MUI: The `value` provided to the Tabs component is invalid.
None of the Tabs' children match with "undefined".
You can provide one of the following values: /tab1, 1, /tab2.

Also, I don't seem to be passing through the index file at all.
src/renderer/src/routes/index.tsx
Do you know why?
I thought I would see the onboarding screens but I am not.

I understand that this is a foundation, but I wonder if it could be cleaned up and clarified a little more before it is merged based on some of the comments below and in here.

src/renderer/src/routes/(MapTabs)/_Map.tsx
ErikSin: Im open to simplfying this. Im just not fully sure how we do that. Do you have any suggestions?

ErikSin:
> Is this related to this issue? [chore-style-tabs](https://github.com/digidem/comapeo-desktop/issues/44)
> 
> I am getting tons of errors / warnings both in the terminal when running tests and the developer console (especially when clicking on the tabs) These relate to invalid props on Dom elements (fullWidth, indicator, selectionFollowsFocus, and textColor) that we are not using explicitly so not sure what is going on here. Did you look into it? It seems like it is something we should try to fix. If for no other reason than that any [other] errors we need to address will be invisible among all the other noise.

Oooh, i didnt open up my console so i didnt see any of these. Will look into them now

src/renderer/src/routes/(MapTabs)/_Map.tsx
ErikSin: Suspense can be as granular as we want. I think there is a balance of having granularity so that only the required things show loaders AND not over complicating the code.

The tab content loads alot of stuff. separate to the map. So wrapping everything in a suspense will cause the map to unmount everytime something loads, even if it is unrelated to the map.

Perhaps this is a personal opinion, but having two suspenses in this component doesnt feel over complicated. Partly because I think we dont need to add anymore suspense throughout the rest of the app because they will be caught by these 2 since it is a layout component. I am open to other thoughts about it if you feel strongly that this is too much

src/renderer/src/routes/(MapTabs)/_Map.tsx
cimigree: Ok thanks

src/renderer/src/routes/(MapTabs)/_Map.tsx
cimigree: Ok cool. I wasn't sure if Suspense was going to have to be used in all components but if this is wrapping everything else than 2 seems reasonable.

src/renderer/src/routes/(MapTabs)/_Map.tsx
ErikSin: THIS BOX IS CAUSING ALL THE CONSOLE ERRORS

src/renderer/src/routes/(MapTabs)/_Map.tsx
cimigree: Maybe a useMapTabs hook to determine what tab to show? But maybe that is unnecessary. Just a suggestion.

src/renderer/src/routes/(MapTabs)/_Map.tab1.tsx
ErikSin: these file names and components are going to change. These are temporary and so I don't think it is necessary to change right now

src/renderer/src/routes/(MapTabs)/_Map.tab2.tsx
ErikSin: See comment above

src/renderer/src/routes/(MapTabs)/_Map.tsx
ErikSin: So, because the tabs needed a `value` prop it had to be aware of the navigation name. This requires the use of `useLocation` hook. This was causing the entire layout screen and all its children (aka the whole app) to rerender on every change in location. So I had to encapsulate the tabs into its own component. I think it simplifies the component?

I think this brings up the question of what is the scope of a component. Gregor has argued that the parent component should contain all the logic, and while it is messier, it does expose everything. And while i don't fully agree, he has some good points. I think there needs to be a more in depth conversation about what are the standards for components.

Regardless, I think the optimization of encapsulating the logic makes sense here because navigation touches everything, and this is a layout component that is the parent for the rest of the app. So the optimization is a necessary as opposed to an over optimization.

ErikSin:
I fixed all the errors. I had to do some hacky things BUT i think these hacky things are better than the alternatives.

1. to deal with the spacing of the tabs, i was originally using a < box/>. This box was causing all the console errors because the tabs was trying to pass it props that didn't exist. So I replaced it with a <Tab /> disabled it, and didnt give it a label or icon. Its techincally in the react tree, but it is not seen or interact-able. I couldn't think of another alternative. The api just doesn't allow for the styling customization that we want
2. the `/` route technically return `null` as the component. But, the `useLayoutEffect` redirects the user to the approriate page. While this is not ideal code, the other alternative is to use the beforeLoading prop in tanstack. But the router then needs use of the hook `useDeviceInfo` to determine where to navigate too. In order to pass this into the router, we need to use a the router provider which needs to initialized on creation of the router. Its quite a lot of complex code for a simple redirect. I think this is a better and cleaner alternative.

src/renderer/src/routes/(MapTabs)/_Map.tsx
```diff
@@ -1,34 +1,99 @@
-import * as React from 'react'
-import { Paper } from '@mui/material'
+import { Suspense } from 'react'
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-desktop/pull/57/commits/a9e9aa0187ce7271e72f7651be25b2a5d2caad7a

src/renderer/src/Theme.ts
```diff
@@ -86,6 +86,7 @@ const theme = createTheme({
 			},
 		},
 	},
+	spacing: 1,
```
ErikSin, line 89: a bunch of spacing was being multipied by 8 because of the way the theme api. So setting the spacing to 1 means any padding or margin is multiplied by one. I don't think we have any utility in one defined default spacing

cimigree: Good find! Sorry about that. I probably added that back in the summer and totally forgot why!

src/renderer/src/routes/__root.tsx
```diff
@@ -3,7 +3,6 @@ import { CssBaseline, ThemeProvider } from '@mui/material'
 import CircularProgress from '@mui/material/CircularProgress'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import { Outlet, createRootRoute } from '@tanstack/react-router'
-import { TanStackRouterDevtools } from '@tanstack/router-devtools'
```
cimigree, line 4: Do you want to remove the actual Devtools also? It is commented out right now.

src/renderer/src/routes/(MapTabs)/_Map.tsx
cimigree: I think it does too. Thanks. I think we will be glad of this later.

cimigree (approved):
Nice work fixing all of those errors! Very impressive!

PR #61, opened 2024-12-13
chore: integrate map libre

### LINKED ISSUE #43: Integrate Map-Libre
Add [Map-Libre](https://maplibre.org/maplibre-gl-js/docs/) as the map rendering library.

cimigree:
closes #43 
- Adds the basic map libre map. I imagine there are specifications somewhere about which map package styler we want to use, but at this juncture, I am not sure what they are. This is just a cartoon globe at this point.
- Adds a location context. I am not sure if this is necessary as I imagine coordinates will come from observations but I am adding it for now
- If we want to use the built in geolocation in production, it looks like we need a Google Cloud Api key, which is why I provided an alternative to try

> GOOGLE_API_KEY
> Geolocation support in Electron requires the use of Google Cloud Platform's geolocation webservice. To enable this feature, acquire a [Google API key](https://developers.google.com/maps/documentation/geolocation/get-api-key) and place the following code in your main process file, before opening any browser windows that will make geolocation requests:
> 
> process.env.GOOGLE_API_KEY = 'YOUR_KEY_HERE'
> 
> By default, a newly generated Google API key may not be allowed to make geolocation requests. To enable the geolocation webservice for your project, enable it through the [API library](https://console.cloud.google.com/apis/library).
> 
> N.B. You will need to add a [Billing Account](https://cloud.google.com/billing/docs/how-to/payment-methods#add_a_payment_method) to the project associated to the API key for the geolocation webservice to work.


package.json
```diff
@@ -60,6 +60,7 @@
 		"electron-is-dev": "^3.0.1",
 		"electron-squirrel-startup": "^1.0.1",
 		"electron-store": "^10.0.0",
+		"maplibre-gl": "^4.7.1",
```
achou11, line 63: although very new, will highlight that v5 was released last week

https://github.com/maplibre/maplibre-gl-js/releases/tag/v5.0.0

although I guess if we use a react wrapper, that should dictate the version we use. most likely v4 for now but if it supports v5, great!

src/renderer/src/components/Map.tsx
achou11, line 1: think it would be worth using a [React wrapper](https://maplibre.org/maplibre-gl-js/docs/plugins/#react) for this given our eventual needs.

https://visgl.github.io/react-maplibre/ seems to be the option I'd lean towards right now

src/renderer/src/contexts/SharedLocationContext.tsx
achou11, line 1: Have we gotten an answer to the question about needing geolocation support in desktop? If there's not an immediate need for it, I would suggest omitting it entirely and have this PR focus on rendering the map with a hardcoded center or (eventually) based on the observation data.

Especially with the caveat you mention about the API key and all, I think introducing is more work than necessary right now.

achou11 (changes requested):

src/renderer/src/components/Map.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const Map = () => {
+	const mapContainerRef = useRef<HTMLDivElement | null>(null)
+	const mapRef = useRef<maplibregl.Map | null>(null)
+	const { location, isLoading } = useSharedLocationContext()
+
+	useEffect(() => {
+		if (!mapContainerRef.current) return
+		const map = new maplibregl.Map({
+			container: mapContainerRef.current,
+			style: 'https://demotiles.maplibre.org/style.json',
```
achou11, line 18: ideally the style that's used comes from `useMapStyleUrl()` in `@comapeo/core-react`. By default, that provides a Mapbox-based map as the default, which I _think_ should work with MapLibreGL but I'm not actually super sure.

package.json
achou11, line 1: let's pin these added deps to exact versions e.g. `npm i -D --save-exact @vis.gl/react-maplibre maplibre-gl`

patches/@comapeo+core+2.1.0+001+fix-smp-fallback-map-setup.patch.patch
achou11, line 1: revert the renaming of this file

src/renderer/src/components/Map.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import { Map as ReactMaplibre } from '@vis.gl/react-maplibre'
+import maplibregl from 'maplibre-gl'
+
+import 'maplibre-gl/dist/maplibre-gl.css'
+
+export function Map() {
```
achou11, line 7: are there any props that might be useful to have to start off with? maybe too early to tell for now 🤔

achou11 (changes requested):

src/renderer/src/components/Map.tsx
```diff
@@ -0,0 +1,36 @@
...
+import maplibregl from 'maplibre-gl'
+
+import 'maplibre-gl/dist/maplibre-gl.css'
+
+import { useMapStyleUrl } from '@comapeo/core-react'
+
+export function Map() {
+	const styleUrlQuery = useMapStyleUrl()
+	const styleUrl =
+		styleUrlQuery.data ?? 'https://demotiles.maplibre.org/style.json'
```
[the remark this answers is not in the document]
achou11: added an issue in core for this: https://github.com/digidem/comapeo-core/issues/982

src/renderer/src/components/Map.tsx
```diff
@@ -0,0 +1,31 @@
...
+			initialViewState={{
+				longitude: center[0],
+				latitude: center[1],
+				zoom: 6,
+			}}
+			dragPan={true}
+			scrollZoom={true}
+			doubleClickZoom={true}
+			style={{ width: '100%', height: '100%' }}
+			mapStyle="https://demotiles.maplibre.org/style.json"
```
achou11, line 22: although I initially suggested using the hook here, it turns out that MapLibre dropped support for `mapbox://` urls, which our default online style currently contains. there's a way of fixing that by using https://github.com/rowanwins/maplibregl-mapbox-request-transformer but it requires the renderer needing access to the the access token, which isn't ideal and probably prone to breaking under other circumstances. Will file an issue to fix the mapbox-specific urls in core but for now, i think it's okay to use the MapLibre demotiles url for the `mapStyle`, like you were doing before.

added an issue in core for this: https://github.com/digidem/comapeo-core/issues/982

patches/@comapeo+core+2.1.0+001+fix-smp-fallback-map-setup.patch.patch
cimigree: Oops

src/renderer/src/components/Map.tsx
cimigree: It might be too early to know as you say.
We can eventually add the onClick for handling observation clicks, or minZoom/maxZoom maybe? As far as I can tell, for right now we only need the basics so I didn't change anything.

achou11 (approved):

src/renderer/src/components/Map.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
```
achou11, line 1: this import isn't necessary

```suggestion
```

PR #59, opened 2024-12-13
chore: add mapeo core react and state manager

ErikSin:
- Adds [CoMapeo-Core/React](https://github.com/digidem/comapeo-core-react)
- Adds Zustand and hook to persist state to local storage through zustand. Follows [the pattern of creating a persisted store and injecting it into React Context](https://tkdodo.eu/blog/zustand-and-react-context). This will allow for easier testing
- Adds a [RouterContext](https://tanstack.com/router/v1/docs/framework/react/guide/router-context) to the router. The context injects the device name and the persisted project id so the appropriate routing can take place. This required a rearrange of the context and providers to be outside of the RouterProvider
- Created an ActiveProjectId store. This is injected deeper into the Router after the user has created an active project. This provider guarantees a projectId saved in persisted state. Otherwise the projectId in persisted state could have been undefined. This simply takes the persisted state, and throws an error if there is no id. otherwise it just propogates to the app

**There seems to be alot of touched files, but many of them are just moving files around**

PR #63, opened 2024-12-15
chore: comapeo core react

ErikSin:
Add comapeo-core/react. Replaces queries with this library, and updated mutations to live in the `hooks/mutation` folder, and to use query keys exposed by comapeo-core/react.

Breaks up the work in #59


cimigree (approved):
Thanks for breaking this apart.

PR #69, opened 2024-12-17
chore(deps): update eslint-related deps

achou11:
Also pins to exact versions to align with our general plan for dependency management in applications


ErikSin (approved):
`@eslint-react/eslint-plugin` is now at 1.20.1 if you want to update that. Otherwise lgtm

PR #73, opened 2024-12-19
chore: add testing environment

ErikSin:
1. Adds React Testing Library and jest dom. configure setup file to import jest dom.
2. Creates a mock client api to allow for testing of @comapeo-core/react
3. Writes test for `/` route and `_Map` route


src/renderer/src/routes/(MapTabs)/_Map.test.tsx
```diff
@@ -1,54 +1,94 @@
...
-	createRouter,
-} from '@tanstack/react-router'
+import { QueryClient } from '@tanstack/react-query'
+import { RouterProvider } from '@tanstack/react-router'
 import { render, screen } from '@testing-library/react'
 import { expect, test } from 'vitest'
 
-import { IntlProvider } from '../../contexts/IntlContext'
-import { MapLayout } from './_Map'
+import { router } from '../../App'
```
achou11, line 8: similar to how we should be working with zustand stores in tests, I _think_ the tests should avoid using this singleton instance and instead create their own instance within the test to avoid potential issues related to shared instances across tests i.e.

```tsx
describe('tabs navigate to correct routes', () => {
	const router = createRouter({ routeTree })

	render(<RouterProvider router={router} />, { wrapper: Wrapper })

    // rest of tests...
```

src/renderer/src/routes/(MapTabs)/_Map.test.tsx
```diff
@@ -1,54 +1,94 @@
...
+		)
+		expect(selectedTabs).toHaveLength(1)
+	})
+
+	test('Tab 1 Screen is showing', async () => {
+		const title = await screen.findByText('Tab 1')
+		expect(title).toBeVisible()
+	})
+
+	test('Clicking "Settings" propogates "/Tab 2" to the navigator', () => {
```
achou11, line 62:
```suggestion
	test('Clicking "Settings" propagates "/Tab 2" to the navigator', () => {
```

src/renderer/src/routes/index.test.tsx
```diff
@@ -0,0 +1,75 @@
+import type { ReactNode } from 'react'
+import { getDeviceInfoQueryKey } from '@comapeo/core-react'
+import { QueryClient } from '@tanstack/react-query'
+import { RouterProvider } from '@tanstack/react-router'
+import { render, waitFor } from '@testing-library/react'
+import { describe, expect, test } from 'vitest'
+
+import { router } from '../App'
```
achou11, line 8: same comment about creating these router instances within the tests instead of using the singleton instance

src/renderer/src/App.tsx
```diff
@@ -26,16 +27,26 @@ const PersistedProjectIdStore = createActiveProjectIdStore({
...
+}: {
+	children: ReactNode
+}) => {
+	return (
+		<ThemeProvider theme={theme}>
+			<CssBaseline />
+			<IntlProvider>{children}</IntlProvider>
+		</ThemeProvider>
+	)
+}
```
achou11, line 52: i don't think abstracting this is necessary for the time being. would suggest just doing duplicate work of setting up the providers in the relevant test helper for now

src/renderer/src/test/helpers/ipc.ts
```diff
@@ -0,0 +1,59 @@
...
+import { MessageChannel } from 'node:worker_threads'
+import { MapeoManager } from '@comapeo/core'
+import {
+	closeMapeoClient,
+	createMapeoClient,
+} from '@comapeo/ipc/dist/client.js'
+import { createMapeoServer } from '@comapeo/ipc/dist/server.js'
+import { KeyManager } from '@mapeo/crypto'
+import Fastify from 'fastify'
+// @ts-expect-error - no types
```
achou11, line 12: if you add the following to the `compilerOptions` in the project root `tsconfig.json`, it should fix this issue:

```
"baseUrl": "./",
"paths": {
    "*": ["node_modules/@digidem/types/vendor/*/index.d.ts"]
}
```

achou11 (changes requested):
Initial feeling is that the actual tests introduced here feel better-suited for being done in the actual app as opposed to the emulated environment. think it's fine for now given that we don't have that set up

Also, would suggest moving the `tests` directory to the renderer root instead of within its `src` directory (see #76 for reasoning) but basically it better follows conventions around node-based projects

PR #74, opened 2024-12-19
chore: create provider to guarantee an active project Id

ErikSin:
Since @comapeo-core/react needs an `ActiveProjectId`, this PR guarantees that there is ActiveProjectId. The `PersistedActiveProjectId` technically has a type of `{projectId:string | undefined}`.  This means that when using the state hook, typescript does not know if the `projectId` is defined.

In order to guarantee a ProjectId, the `_Map` layout component, which wraps all the routes after onboarding, does a beforeLoad check to see if projectId exists, and redirects user back to onboarding if not.

This `_Map` layout component then wraps everything in the `GuaranteedActiveProjectIdProvider` which propgates this `ActiveProjectId`, and throws an error if no project Id, allowing for a typesafe `ActiveProjectId` that is always defined.

Waiting on #73 to be merged to create test


ErikSin:
@gmaclennan and @achou11 I wanted to get your opinion on how we should typesafe the `activeProjectId` since we are planning on using it often for @comapeo-core/react

gmaclennan (changes requested):
I'd like to return to the suggestion to put the project ID in the route state (e.g. part of the URL). When we are storing the active project ID separately to the route state, we have some routes where it must exist, but the two states can be out of sync (e.g. the route state/URL can be pointing to a page that requires the project ID, but the active project ID is not set). This creates the complications that you are trying to deal with in this PR, which doesn't entirely solve the problem, and creates additional complexity in the app. If the project ID is in the route state, then it is impossible to be on a route where the project ID is required, but it is undefined - if the project ID is missing from the route state, then the route is not found.

You have additional states as well to consider that should not be possible, but could happen, e.g. no active project ID, but a project is already created / exists on the backend.

I suggest that we continue with our architecture of an index route that redirects:

1. No device name => redirect to onboarding to add device name
2. No projects in the backend (`useManyProjects().length === 0`) => redirect to createOrJoinProject route
3. Projects exist in backend, redirect the oldest/newest (should only be one, but just in case) e.g. `/${projectId}/observations/`

When we want to support multi-project, we can remove (3) from the index route, and instead show a list of projects to choose from.

We can then handle missing projects, or any other missing resource like an observation, with a Not Found route (since we are using React Query, rather than tanStack query data loading, this would be a special case of the ErrorBoundary component), where the user can redirect to `/` from the project not found screen, or `/${projectId}/observations/` from the observation not found screen.

Then for persistence, we can decide whether we want to persist the entire URL, or just the projectId. We can do that outside of "react world" to make things a lot simpler for us:

```ts
const router = createRouter(routeTree)
const { pathname, search, hash } = JSON.parse(localStorage.getItem("location"))
router.navigate({ pathname, search, hash, replace: true })
router.subscribe("onResolved", () => {
  const { pathname, search, hash } = window.location
  localStorage.setItem("location", JSON.stringify({ pathname, search, hash }))
})
```

Can be abstracted if we want to keep App.tsx more readable:

```ts
function persistRouterState(router) {
  const { pathname, search, hash } = JSON.parse(localStorage.getItem("location"))
  router.navigate({ pathname, search, hash, replace: true })
  router.subscribe("onResolved", () => {
    const { pathname, search, hash } = window.location
    localStorage.setItem("location", JSON.stringify({ pathname, search, hash }))
  })
}

const router = createRouter(routeTree)
persistRouterState(router)
```

We could then change persistRouterState as needed down the line according to user requirements, e.g. to only store the project ID from the route if we want, or do any other logic we want around what is persisted in terms of the users navigation state (e.g. might want to not persist nested navigation of individual images, or routes editing something)


If you really want to continue with active project ID stored separately to the router state, then there are a few more changes needed here to make it work. I would avoid having a separate provider, and instead have the `useActiveProjectId()` hook throw if there is not a value. An extra provider that is not actually storing it's own state, but instead contains derived state from the ActiveProjectId provider, is kind of confusing. I'd really recommend avoiding the pattern of active project ID not in the router state (url) though, because it's necessary for many routes and having it not part of the route state is going to require wrappers in several places, and separate handling of "project not found" vs. "no active project id".

PR #79, opened 2025-01-06
implement ipc for file selection

achou11:
Related to #55

Sets up the mechanisms for initiating the selection of a file in the filesystem from the renderer side and retrieving the full path to the selected file. Having this will enable features such as importing a project config file or custom map file.

This PR introduces the following:

- IPC handler on the main window for summoning the open file dialog and returning the full path to the selected file (or `undefined` if no file is selected). Multi-file support was considered but we don't have an immediate need for it, so opted to refrain from including in this implementation.
- new renderer-available "runtime" method `selectFile(extensionFilters)` which sends the relevant IPC and waits for the result.
- mutation hook `useSelectProjectConfigFile()` which is unused in this PR, but will help #55. Example usage could look like the following:

    ```tsx
    onClick={() => {
	    selectConfigFile.mutate(undefined, {
		    onSuccess: (filePath) => {
			    if (filePath) {
				    // Do something with selected file...
			    } else {
				    // No file selected due to dialog cancellation
			    }
		    },
		    onError: (error) => {
			    // Handle error...
		    },
	    })
    }}
    ```

### DIFF
--- a/src/main/app.js
+++ b/src/main/app.js
@@ -6,12 +6,14 @@ import {
 	BrowserWindow,
 	MessageChannelMain,
 	app,
+	dialog,
 	ipcMain,
 	safeStorage,
 	utilityProcess,
 } from 'electron/main'
 
 import { Intl, getSystemLocale } from './intl.js'
+import { APP_IPC_EVENT_TO_PARAMS_PARSER } from './ipc.js'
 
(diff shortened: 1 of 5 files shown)


cimigree (approved):
Looks great. Thanks for doing this. I tested in the electron console and it worked as expected.

PR #80, opened 2025-01-07
chore(deps): update deps related to tooling

achou11:
Shouldn't have any effect on the running application. Planning to do various follow ups for deps that could affect that.

For all of the affected deps in this PR, I've also pinned them to exact versions, to align with our decision to do that with our  deps moving forward.


cimigree (approved):
I ran the code and looked at changelogs. Seems like an innocuous update.

PR #81, opened 2025-01-07
feat: update runtime.selectFile() to also return the file name

achou11:
Updates `runtime.selectFile()` to return an object that contains the path and name of the selected file, as opposed to just the path. Informed by the needs of #55

### DIFF
--- a/src/preload/main-window.js
+++ b/src/preload/main-window.js
@@ -28,16 +28,36 @@ const runtimeApi = {
 
 	// Files
 	async selectFile(extensionFilters) {
-		const filePath = await ipcRenderer.invoke('files:select', {
+		/** @type {unknown} */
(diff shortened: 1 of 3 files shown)


src/preload/main-window.js
```diff
@@ -28,16 +28,36 @@ const runtimeApi = {
...
+	}
+
+	if (typeof value.path !== 'string') {
+		throw new Error('Value has invalid path field')
+	}
+
+	if (typeof value.name !== 'string') {
+		throw new Error('Value has invalid name field')
+	}
+}
```
achou11, line 61: ideally would use valibot for this but I don't have bundling set up for the preload just yet so I think it would require some additional tooling work.

cimigree (approved):
Thanks.

PR #82, opened 2025-01-08
chore(deps): update comapeo-related deps

achou11:
Updates and pins them to exact versions. There shouldn't be any breaking changes that affect the app right now.


cimigree (approved):
Does the patch file name need to be updated for @comapeo/core also?
I get this warning when I install node_modules:
```
Warning: patch-package detected a patch file version mismatch

  Don't worry! This is probably fine. The patch was still applied
  successfully. Here's the deets:

  Patch file created for

    @comapeo/core@2.1.0

  applied to

    @comapeo/core@2.3.0

  At path

    node_modules/@comapeo/core

  This warning is just to give you a heads-up. There is a small chance of
  breakage even though the patch was applied successfully. Make sure the package
  still behaves like you expect (you wrote tests, right?) and then run

    patch-package @comapeo/core

  to update the version in the patch file name and make this warning go away.
```

achou11:
> Does the patch file name need to be updated for @comapeo/core also?

ah yeah forgot that. done via 13b043e

Issue #92, opened 2025-05-21
Desktop MVP - Update UI - Onboarding/Project Joining


ximenabb:
**Description**
As we have updated CoMapeo Mobile UI, we are updated Desktop to match mobile. We will be providing the designs in batches, prioritizing the elements that don't require mapping with others (project invite, exchange).

Feature spec :  https://www.notion.so/digidem/CoMapeo-Desktop-MVP-Updated-1fa1b08162d580fe964fd30f12a7fa3d?pvs=4

Clickable prototype : https://www.figma.com/proto/UIA00RSP8Sz24V0vXvQHeL/-Desktop--CoMapeo?node-id=1162-2189&t=GuvTh8j7xljUE7Kj-1&scaling=min-zoom&content-scaling=fixed&page-id=8%3A2&starting-point-node-id=1162%3A2189&show-proto-sidebar=1

Design files : https://www.figma.com/design/UIA00RSP8Sz24V0vXvQHeL/-Desktop--CoMapeo?node-id=1162-2189&t=WPii7YIUdCjLyl5s-0

**To Do**

## Onboarding flow
<img width="500" alt="Image" src="https://github.com/user-attachments/assets/ad66faa5-dd5b-4838-902a-34eca21b5513" />

**Welcome** 
- [x] Update copy as seen on figma
- [x] When tapping on "Get started", go to Data & Privacy screen

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/4391e737-66da-4b10-8aba-99dcd783d04a" />

**Data & Privacy**
- [ ] Update copy as seen on figma
- [ ] When tapping on "Learn More", go to "Learn More about Data & Privacy" svreen
- [ ] When tapping on "Next" go to "Name Device" screen

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/a8acac55-cf55-41ce-a36d-48dbe7eb2fc5" />

**Learn More about Data & Privacy**
- [ ] Same Learn More as designed in prototype
  - Updated designs: https://www.figma.com/design/UIA00RSP8Sz24V0vXvQHeL/-Desktop--CoMapeo?node-id=1632-3458&t=Y3lKrfSvxXNqQTRR-0

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/7e4ce3fe-a075-4719-9c08-f48613ac06c4" />

**Name Device**
- [ ] Update copy as seen on figma
[...]

PR #109, opened 2025-06-18
update UI for initial onboarding screen

achou11:
Towards #92 

Other notable changes:

- adjusts initial window size
- adjusts minimum width and height for window
- enables [strict mode](https://react.dev/reference/react/StrictMode)
- a variety of fixes and foundational changes related to the theming setup. this will very likely introduce UI issues for other screens, but those will be addressed as follow-ups.
 
 
 ---
 
- Full screen:
 
    image

- Minimum width:

    image

### DIFF
--- a/src/renderer/src/Theme.ts
+++ b/src/renderer/src/Theme.ts
@@ -1,4 +1,4 @@
-import { createTheme } from '@mui/material/styles'
+import { createTheme, responsiveFontSizes } from '@mui/material/styles'
 
 import {
 	ALMOST_BLACK,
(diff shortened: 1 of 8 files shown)


src/renderer/src/Theme.ts
```diff
@@ -52,11 +76,21 @@ const theme = createTheme({
...
+			fontSize: '1.5rem',
+			fontWeight: 500,
+			lineHeight: 1.25,
+		},
 	},
 	palette: {
 		text: {
 			primary: ALMOST_BLACK,
 			secondary: DARK_GREY,
+			inverted: WHITE,
```
achou11, line 93: kind of hacky but works okay for now

src/renderer/src/contexts/IntlContext.tsx
achou11, line 1: using the text mappings seems to surface an key uniqueness issue caused by something internal to react-intl. Tried using https://formatjs.github.io/docs/react-intl/components/#wraprichtextchunksinfragment but didn't seem to work. Might need to revisit later but don't think it's an issue right now.

src/renderer/src/routes/Onboarding/index.tsx
```diff
@@ -41,194 +51,120 @@ const m = defineMessages({
...
+										<img src={WORLD_MAP_IMAGE_URL} height={30} width={30} />
+										<ListItemText
+											slotProps={{ primary: { color: 'textInverted' } }}
+										>
+											{t(m.mapAnywhere)}
+										</ListItemText>
+									</ListItem>
+									<ListItem sx={{ gap: 5 }}>
+										{/* TODO: Replace with correct image */}
+										<img src={RAISED_HAND_IMAGE_URL} height={30} width={30} />
```
achou11, line 128: need to get the appropriate asset for this. will do as a follow-up

src/renderer/src/routes/Onboarding/index.tsx
```diff
@@ -1,36 +1,46 @@
...
-import { Text } from '../../components/Text'
-import TopoBackground from '../../images/TopoLogo.svg'
-import Calling from '../../images/calling.png'
-import LockedWithKey from '../../images/closed_lock_with_key.png'
-import RaisedFistMediumSkinTone from '../../images/raised_fist_medium_skin_tone.png'
-import WorldMap from '../../images/world_map.png'
+import { BLACK, BLUE_GREY, DARK_COMAPEO_BLUE } from '../../colors'
+import { ButtonLink } from '../../components/button-link'
+// TODO: Remove `?url` after getting rid of vite-plugin-svgr
+import TOPO_IMAGE_URL from '../../images/TopoLogo.svg?url'
```
achou11, line 17: highlighting for future reference. without the `?url`, this import gets processed by the svgr plugin and the import becomes a react component, which is not desired in this case especially.

Issue #91, opened 2025-04-23
Consider bundling main process code when packaging


achou11:
Right now we don't do any processing of the code that runs in main process i.e. anything that exists in `src/main/` and `src/services` directories. The benefit of this is that the code we write in development is basically the code that runs when the app is packaged. However, the main drawback is that it requires dependency code that's used to exist as-is within the distributable when packaged, which dramatically increases the size of our distributable (100+ mb uncompressed).

On mobile this is a more pressing issue due to device constraints so we have a build process that bundles the relevant node-based code and do some extra steps to make sure that certain modules/directories are copied over based on the app's needs. We can introduce something similar for desktop, which uses similar tooling under the hood (rollup).

Things to consider:

- I think we'd have to bundle code in both `src/main/` and `src/services/` (can't commit to just the former, for example). I think bundling services may be tricky because the main process uses the services via utility processes. For example, [electron-vite](https://electron-vite.org/guide/dev#utility-process-and-child-process) has an internal plugin to facilitate this when bundling with vite.

- It would be good to see how much of a reduction in distributable size doing this work results in. It may be that introducing the additional build steps and all may not be worth it if we're interested in the change in app size. There are other benefits to bundling such as runtime performance improvements, but not sure how much of a bottleneck that is to begin with for us.

PR #190, opened 2025-08-05
chore: release automations

### LINKED ISSUE #97: Set up release automations
I'm guessing we'll want something similar to what we do for mobile (documented [here](https://github.com/digidem/comapeo-mobile/blob/develop/docs/Releases.md)). Involves porting the relevant repo rulesets and introducing relevant GitHub workflows.

The main difference will be how to handle built artifacts and their distribution. For distribution, it'll likely be a mix of relying on GitHub's artifact storage (for RC builds) and GitHub releases (for release builds).

For official releases, there's also the additional factor of the Cloudflare integration that's used to add a link to the website that points to the latest releases for download. Probably something we can figure out separately and further down the road.

achou11:
Closes #97 

Ports over a good chunk of the automations that come from mobile, but there are also some notable distinctions that I want to point out. There's a decent chance that some of these differences aren't necessary.

1. ~Production releases are solely triggered based on release tags being created/pushed in the repo. See `tag-release.yml` and `build-production.yml`. The main reasoning for this is to ensure that the tag exists prior to creating the GitHub release. Could be misunderstanding how [Github's API](https://docs.github.com/en/rest/releases/releases#create-a-release--parameters) works, but seems like the tag needs to exist when creating a release.~ EDIT: No longer relevant

2. The comment that's created after merging a RC PR points to the relevant GitHub release, instead of some build URL.

3. ~The comment that's created when invoking a RC build for a RC PR does not provide a URL that points to the relevant assets. This is not intentional, I just haven't figured out the best approach for this, as there are multiple asset URLs to provide since we run the application build step using a matrix strategy (one per relevant target platform).~ EDIT: I think I figured out a sufficient solution. See https://github.com/digidem/comapeo-desktop/pull/190/commits/d25437518831c3a7838b3f86376e88ba1bd8e61e.

4. No changelog-specific steps right now. However, when we create a release, we currently have it set up such that GitHub will generate a changelog based on the commits that have happened between releases. I have no idea how well this is done but wanted to give it a try to see if it could be useful.

Other important notes:

1. The creation of a release means creating a draft release for the repo and uploading the relevant assets to it. As currently implemented, it still requires someone to officially press something to actually publish it.

2. Currently uses [`@electron-forge/publisher-github`](https://github.com/electron/forge/tree/main/packages/publisher/github) for creating GitHub releases. ~Had to include a module patch that makes it so that it uses our custom version format in order to do tag/release resolution. Otherwise it attempts to use the version from the package.json file.~ EDIT: Better approach used based on suggestion from Gregor

3. Currently does not implement a publisher related to something like S3 or Cloudflare. Might leave that as a follow-up task depending on the amount of work needed for it.

Remaining items:

- [x] Figure out what URL to provide in the RC build PR comment
- [x] Port over + write release docs

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,6 @@
 # Summary:
 #
-# Creates a release tag on the release branch when a release candidate PR is merged.
+# Creates a release based on the release branch when a release candidate PR is merged.
 #
 # Triggers:
 #
@@ -8,9 +8,8 @@
 #
 # Effects:
 #
-# - Creates a release tag to the release branch named `v${version}` using the merge commit in the targeted `release/**` branch,
-#   where `version` is `<minor>.<patch>` e.g. `1.2.0` is truncated to `2.0`.
-# - Triggers a production release that creates the builds, creates a (draft) GitHub release, and publishes the assets to that release.
+# - Creates a draft GitHub release
+# - Creates application builds and uploads them to the GitHub release.
 # - Comments on the the release PR pointing to the GitHub release URL.
 name: Release
 
@@ -22,12 +21,10 @@ on:
       - 'release/**'
 
 jobs:
-  tag:
-    name: Create tag
+  create_release:
+    name: Create release
     if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
-    outputs:
-      version_tag: v${{ steps.version.outputs.release_version_short }}
     steps:
       - name: Create GitHub App Token
         uses: actions/create-github-app-token@v1
@@ -48,30 +45,11 @@ jobs:
         id: version
         uses: ./.github/parse-version
 
-      - name: Tag release
-        run: |
-          git config --global user.name 'awana-release-bot[bot]'
-          git config --global user.email '${{ vars.RELEASE_BOT_USER_ID }}+awana-release-bot[bot]@users.noreply.github.com'
-          git tag ${{ steps.version.outputs.releaseShort }}
-          git push origin git tag ${{ steps.version.outputs.releaseShort }}
-
-  create_release:
-    name: Create release
-    needs: tag
-    runs-on: ubuntu-latest
-    steps:
-      - name: Create GitHub App Token
-        uses: actions/create-github-app-token@v1
-        id: app-token
-        with:
-          app-id: ${{ vars.RELEASE_BOT_APP_ID }}
-          private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
-
       - name: Create release on GitHub
         env:
           GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
           REPOSITORY: ${{ github.repository }}
-          TAG_NAME: ${{ needs.tag.outputs.version_tag }}
+          TAG_NAME: ${{ steps.version.outputs.release_version_short }}
         run: |
           release_url=$(gh release create \
             --repo "$REPOSITORY" \
@@ -82,7 +60,7 @@ jobs:
 
   build:
     name: Create builds
-    needs: tag
+    needs: create_release
     uses: ./.github/workflows/create-builds.yml
     with:
       app_type: 'production'
@@ -100,10 +78,6 @@ jobs:
         with:
           app-id: ${{ vars.RELEASE_BOT_APP_ID }}
           private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
-
-      - name: Parse version
-        id: version
-        uses: ./.github/parse-version
 
       - name: Comment on PR
         env:


.github/parse-version/action.yml
achou11, line 1: same as mobile

.github/workflows/build-bot.yml
achou11, line 1: same as mobile

.github/workflows/build-production.yml
achou11, line 1: noting that this is what is responsible for triggering the creation of production releases

.github/workflows/build-rc.yml
```diff
@@ -0,0 +1,76 @@
...
+          app-id: ${{ vars.RELEASE_BOT_APP_ID }}
+          private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
+
+      - name: Comment on PR
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          PR_NUMBER: ${{ inputs.pr_number || github.event.pull_request.number }}
+          # TODO: Add link to build job's artifact URL.
+          COMMENT_BODY: |
+            :construction: Release candidate build based on ${{ env.BUILD_SHA }} (potential merge of ${{ env.PR_HEAD_SHA }})
```
achou11, line 73: could use some help/ideas with this. the problem is that the `create-builds.yml` workflow runs a matrix, so getting a single URL as the output is not trivial. Could potentially just link to every artifact URL, which is probably feasible but haven't thought much about how to actually implement.

.github/workflows/check-pr-to-release.yml
achou11, line 1: same as mobile

.github/workflows/create-builds.yml
```diff
@@ -0,0 +1,112 @@
...
+        type: string
+  workflow_dispatch:
+    inputs:
+      app_type:
+        description: 'Type of application. Valid values are: internal, or release-candidate'
+        required: true
+        type: choice
+        # We only allow internal and release candidate builds when doing manual dispatches.
+        # Production builds should only be triggered by the appropriate calling workflows.
+        options: ['internal', 'release-candidate']
```
achou11, line 27: I was thinking that allowing manual dispatch for these will be useful. The only question I have is whether production builds should also allowed to be manually dispatched. My intuition says no, but wondering how to handle cases where the production release build workflow fails for some reason. For example:

1. Merge RC PR -> create release tag
2. New release tag -> trigger production release
3. Production release fails -> ???

What's the course of action in (3)? Re-run the failed job? Maybe that's sufficient.

.github/workflows/create-builds.yml
```diff
@@ -0,0 +1,112 @@
...
+          [
+            { name: 'linux', image: 'ubuntu-latest' },
+            { name: 'macos', image: 'macos-latest' },
+            { name: 'macos-intel', image: 'macos-13' },
+            { name: 'windows', image: 'windows-latest' },
+          ]
+    runs-on: ${{ matrix.os.image }}
+    outputs:
+      artifact_id: ${{ steps.artifact.outputs.artifact-id }}
+      artifact_url: ${{ steps.artifact.outputs.artifact-url }}
```
achou11, line 44: these are not used at the moment, as they are not really reflective of what we want to provide as an output. Since this workflow runs in a matrix, these outputs will only reflect the value coming from the last-run iteration of the matrix.

What we really want is either:

1. A list of all of the artifact outputs
2. A URL that only points to the general artifacts page i.e. `https://github.com/example-org/example-repo/actions/runs/1/artifacts/`

.github/workflows/create-builds.yml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Create distributables
+        run: npm run forge:make
+
+      - name: Upload artifacts
+        id: artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.os.name }}-${{ inputs.build_sha }}
+          compression-level: 0
```
achou11, line 100: i _think_ this is desired? Could be wrong though: https://github.com/actions/upload-artifact/tree/main?tab=readme-ov-file#altering-compressions-level-speed-v-size

.github/workflows/create-builds.yml
```diff
@@ -0,0 +1,112 @@
...
+          path: |
+            out/make/**/*.AppImage
+            out/make/**/*.dmg
+            out/make/**/*.exe
+            out/make/**/*.zip
```
achou11, line 105: Think I need a more thorough set of patterns. For example, seems like the squirrel maker produces a few files: https://www.electronforge.io/config/makers/squirrel.windows

.github/workflows/create-builds.yml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Publish application
+        if: ${{ inputs.APP_TYPE == 'production' }}
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: |
+          npm run forge:publish
```
achou11, line 112: think the ideal would be that this step occurs from the calling workflow as opposed to being conditionally run in this workflow. however, this forge command needs the assets produced by the `npm run forge:make` command in an earlier step, hence its placement here.

maybe there's a clean way of achieving what I want, but I lack the knowledge right now...

.github/workflows/hotfix-release.yml
achou11, line 1: Same as mobile

.github/workflows/post-release-check.yml
achou11, line 1: Same as mobile

.github/workflows/scheduled-release.yml
achou11, line 1: Same as mobile

.github/workflows/tag-release.yml
achou11, line 1: Basically the same as [`build-release.yml` in mobile](https://github.com/digidem/comapeo-mobile/blob/develop/.github/workflows/build-release.yml) but without the attempt to create the build itself. See PR description for reasoning.

forge.config.js
```diff
@@ -204,37 +223,37 @@ const plugins = [
...
 		new MakerRpm({ options: { icon: './assets/icon.png' } }, ['linux']),
 	],
+	publishers: [
+		new PublisherGithub({
+			draft: true,
+			force: true,
+			generateReleaseNotes: true,
+			repository: { owner: 'digidem', name: 'comapeo-desktop' },
+			tagPrefix: 'v',
+		}),
```
achou11, line 256: as noted in the PR description, we have this set up to create release drafts that:

1. require human intervention to formally publish
2. generates a changelog in the release body text.

.github/workflows/build-rc.yml
achou11: Think the solution I'm using in https://github.com/digidem/comapeo-desktop/pull/190/commits/d25437518831c3a7838b3f86376e88ba1bd8e61e is sufficient. Open to alternative ideas.

.github/workflows/create-builds.yml
achou11: This is no longer relevant, but still am curious about potential ways of achieving something like this, for future reference!

gmaclennan:
It's not necessary to create the tag before creating the release. Creating the release will create the tag automatically. I validated this by creating a test repo and creating a release via the API:

```
❯ gh api \
  --method POST \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  /repos/gmaclennan/release-test/releases \
   -f 'tag_name=v1.1' -f 'target_commitish=main' -f 'name=v1.1' -f 'body=Description of the release' -F "draft=false" -F "prerelease=false" -F "generate_release_notes=false"
```

I think it's best to avoid the tag-creation workflow because it adds a bit of indirection and the possibility of creating tags without sign-off via the PR (but could be avoided by creating repo rules for tag creation)

You shouldn't need to patch the publisher, I think you can use the [`readPackageJson` hook](https://www.electronforge.io/config/hooks#readpackagejson)

achou11:
@gmaclennan added some follow-ups based on feedback:

- using the `readPackageJson` hook instead of having a module patch for the version: https://github.com/digidem/comapeo-desktop/pull/190/commits/1a1cd3f4eef982f374642b59716fad4279d2e09b. I only enable this when creating non-development builds. 

- consolidated the workflows in question in https://github.com/digidem/comapeo-desktop/pull/190/commits/ea0763013576485198c29a698e6b692e27634fbc. One thing to note is that I decided to do the tagging job first, followed by the build and comment jobs. My reasoning is to avoid relying on the forge publisher plugin to create the release tag, which felt a bit too hidden for my liking.

.github/workflows/create-builds.yml
```diff
@@ -0,0 +1,116 @@
...
+            out/make/**/*.zip
+
+      - name: Publish application
+        if: ${{ inputs.APP_TYPE == 'production' }}
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: |
+          git config --global user.name 'awana-release-bot[bot]'
+          git config --global user.email '${{ vars.RELEASE_BOT_USER_ID }}+awana-release-bot[bot]@users.noreply.github.com'
+          npm run forge:publish
```
achou11, line 116: I'm a bit unsure about whether the git config steps are needed prior to running the publish command. Not sure how attribution to the creation of a release is done.

achou11:
Marked the PR as ready for review but still need to include some docs for the release process.

.github/workflows/hotfix-release.yml
```diff
@@ -0,0 +1,123 @@
...
+jobs:
+  check-inputs:
+    name: Check Inputs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check release branch exists
+        # Check inputs with a GitHub API request rather than checking out repo
+        # first and using git show-ref --verify, because this is much faster.
+        run: |
+          curl --location \
```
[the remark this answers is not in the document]
achou11: yeah this is just copied over from mobile, which is copied from the release automation repo.

as mentioned in other spaces, it'd probably make more sense for this to just directly use the workflow in the release automations repo

.github/workflows/create-builds.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Create env file
+        env:
+          APP_TYPE: ${{ inputs.app_type }}
+        run: |
+          echo "APP_TYPE=${{ env.APP_TYPE }}" >> .env
+          echo "ONLINE_STYLE_URL=${{ secrets.ONLINE_STYLE_URL }}" >> .env
+
+          if [[ ${{ env.APP_TYPE }} != 'production' ]]; then
+            echo "VITE_FEATURE_TEST_DATA_UI=true" >> .env # We enable the test data creation UI for non-production builds
+          fi
```
gmaclennan, line 86: This seems very round-about? I thought the reason for using environment variables is to be able to set them in the environment.

.github/workflows/create-builds.yml
gmaclennan: I think the speed that this action takes is inconsequential, but also space saving isn't that relevant because we're on a sponsored plan and don't pay for artifact storage.

gmaclennan (commented):
a few small tweaks, but generally looks good. The whole environment variable management is confusing to me (although environment variables are confusing in general - see Expo as an example). If they are for build-time configuration, then I think they should be environment variables, using dotenv for convenience when developing so that a dev can just write them to a file, but setting them in the environment during build. There are two types of build-time environment variables: those that affect build steps, and those that change the runtime code that's built via bundlers etc. Additionally there is app configuration, which does not make sense to me to use as an environment variable. There is the slight question about "secrets" which you don't want to checkin to the repo, although particularly with a desktop app, they very much are not "secret" so not sure what we do with our mapbox token for example.

.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          token: ${{ steps.app-token.outputs.token }}
+
+      - name: Parse version
+        id: version
+        uses: ./.github/parse-version
+
+      - name: Tag release
```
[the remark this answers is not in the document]
achou11: yeah i think it'd be worth considering. i'm trying to not overly rely on the publisher plugin so whatever work we can save it, i think the better.

.github/workflows/create-builds.yml
achou11: Thinking about dropping (3) as I look through what the main process is actually using those env variables for (starts [here](https://github.com/digidem/comapeo-desktop/blob/6d106db427b5dce6007ee5276e58693778b110fa/src/main/utils.js#L13-L19)). It probably doesn't make sense to have a file that defines those values at runtime now that I think about it.

achou11: i think the `APP_TYPE` is also necessary for the main process to know at runtime, as it will be used for when we set up Sentry.

achou11: I think for now, I'm going to keep this step. I can update the step name to be clearer about why we're creating an env file here.

gmaclennan: I think the issue with packaging a .env in the app to read configuration values is that a dev could easily put a sensitive value in there and not expect it to ship publicly. I think it's better to either bundle backend code and use a transform to replace selected env variables; add an app.config.json to source control and package with the app, or generate the app.config.json at build time based on env variables. I just think mixing the two is confusing as it's doing two separate things.

gmaclennan: I think the expo approach is pretty solid: a dynamic app.config.js which can read environment variables and writes the config to files that are not checked in, but are included at runtime (all-be-it in the expo case it's much more complicated that just a json - eg manifest.xml etc)

achou11: > I think the issue with packaging a .env in the app to read configuration values is that a dev could easily put a sensitive value in there and not expect it to ship publicly.
> ...
>  I just think mixing the two is confusing as it's doing two separate things.

Yep both very true...

> I think it's better to either bundle backend code and use a transform to replace selected env variables

Yeah I think this is our best long-term option. Already have an issue for it here: #91 

> generate the app.config.json at build time based on env variables.

Think I'll opt for this approach for now.

.github/workflows/release.yml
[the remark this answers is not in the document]
achou11: done via 3040ea3e47282eca37d1735986160e5f13d6cb8d

.github/workflows/release.yml
```diff
@@ -0,0 +1,119 @@
...
+        run: |
+          release_url=$(gh release create \
+            --repo "$REPOSITORY" \
+            --draft \
+            --verify-tag \
+            --generate-notes \
+            --title "$TAG_NAME" \
+            "$TAG_NAME" | jq '.url' --raw-output)
+
+          echo "release_url=$release_url" >> $GITHUB_OUTPUT
```
achou11, line 84: could use another set of eyes on this to make sure it's correct.

https://cli.github.com/manual/gh_release_create

https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release

.github/workflows/create-builds.yml
achou11: Removed this env file creation step via https://github.com/digidem/comapeo-desktop/pull/190/commits/fab883cf192488e2ba70e17005b6ac218a431bc8.

Planning to do the creation of the `app.config.json` that makes use of the env variables as a follow-up

achou11 (commented):
Addressed some additional feedback and also added preliminary docs (will add screenshots once this PR is merged).

.github/workflows/create-builds.yml
achou11: For posterity, stacked PR is here: https://github.com/digidem/comapeo-desktop/pull/191

gmaclennan (approved):
Looks good, just a couple of small changes then good to go I think.

.github/workflows/release.yml
```diff
@@ -0,0 +1,119 @@
...
+          token: ${{ steps.app-token.outputs.token }}
+          sparse-checkout: |
+            package.json
+          sparse-checkout-cone-mode: false
+
+      - name: Parse version
+        id: version
+        uses: ./.github/parse-version
+
+      - name: Tag release
```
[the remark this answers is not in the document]
achou11: ah good point, definitely would like to be resilient to dangling tags. so is the suggestion to basically remove this tag step and just do the release step and let that (eventually) handle the creation of the tag?

.github/workflows/release.yml
```diff
@@ -0,0 +1,119 @@
...
+      - name: Create GitHub App Token
+        uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ vars.RELEASE_BOT_APP_ID }}
+          private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
+
+      - name: Parse version
+        id: version
+        uses: ./.github/parse-version
```
gmaclennan, line 109: don't think you need this any more

.github/workflows/release.yml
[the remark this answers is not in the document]
gmaclennan: yes just merge this job with the create release job (to do the parse version before creating the release so that you have the release name)

.github/workflows/release.yml
achou11: addressed via https://github.com/digidem/comapeo-desktop/pull/190/commits/99c1fd20d71abf2c61a1352a58c09620aeea8cc0

.github/workflows/release.yml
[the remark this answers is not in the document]
achou11: done via https://github.com/digidem/comapeo-desktop/pull/190/commits/a01e193b0d12b105c31079e096fcf9a887943aae

.github/workflows/create-builds.yml
```diff
@@ -0,0 +1,109 @@
...
+            out/make/**/*.dmg
+            out/make/**/*.nupkg
+            out/make/**/*.rpm
+            out/make/**/*.zip
+
+      - name: Publish application
+        if: ${{ inputs.APP_TYPE == 'production' }}
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: npm run forge:publish
```
gmaclennan, line 109: kind of feels like this should be in release.yml, because it depends on the release creation in that step (the contract to not run this workflow outside of release.yml for app_type production feels very loose). However that can be a follow up - might be tricky to pass the correct artifacts etc. You could make the create build an action like parse-version so it can be run in a single step. Anyway, fine like this for now.

PR #200, opened 2025-08-12
chore: set up Sentry

### LINKED ISSUE #101: Improved observability with Sentry
Umbrella issue documenting our desired integration of Sentry, similar to what's documented in https://github.com/digidem/comapeo-mobile/issues/1208

- Need to create project for the desktop application on Sentry

- Integrate with main process and services using https://docs.sentry.io/platforms/javascript/guides/electron/

- Integrate with renderer process using https://docs.sentry.io/platforms/javascript/guides/react/

- Environments we want:

  - Development

  - QA

  - Production

### LINKED ISSUE #102: Sentry integration backend
The "backend" currently consists of:

1. Electron's "main" process

2. Any "services" that are run using Electron's [utilityProcess](https://www.electronjs.org/docs/latest/api/utility-process) API. These are essentially NodeJS child processes.

### Known Tasks:

- [ ] Create Sentry project that's specific to the backend for CoMapeo Desktop

- [ ] Set up Sentry for the main process using their [Electron SDK](https://docs.sentry.io/platforms/javascript/guides/electron).

### Questions

- Does the Sentry [Node SDK](https://docs.sentry.io/platforms/javascript/guides/node/) still need to be used in this case? Mostly relevant in the case of the "core service", which is where we run the core-related backend.

### LINKED ISSUE #103: Sentry integration (frontend)
Currently, we only have the one renderer process that the application uses (the main React application).

### Known tasks

- [ ] Create a Sentry project specific to the CoMapeo Desktop frontned
- [ ] Set up Sentry's [Electron SDK](docs.sentry.io/platforms/javascript/guides/electron/).
  - docs for setup when using another framework-specific SDK: https://docs.sentry.io/platforms/javascript/guides/electron/#framework-specific-sdks
  - docs for offline support: docs.sentry.io/platforms/javascript/guides/electron/#offline-support
- [ ] Set up Sentry's [React SDK](https://docs.sentry.io/platforms/javascript/guides/react/).

### Questions

achou11:
Closes #101 
Closes #102
Closes #103

Uses Sentry's Electron, React, and Vite integrations.

- I've added the following repository secrets: `SENTRY_AUTH_TOKEN`, `SENTRY_ORG`, and `SENTRY_PROJECT`.

- I've updated the config store to hold information about the Sentry user that gets rotated approximately every month (assuming the app is opened within that span). Mostly adapted the logic from mobile for this.

- Updated the app to send failed mutations as captured exceptions. There's probably still more work to do related to error boundaries or handling query failures, but will leave that as a follow-up.

- Have a couple of uncertainties around the init options that I'm using (in both main and renderer). Would be helpful to get some eyes on that specifically.


src/main/index.js
```diff
@@ -68,14 +73,52 @@ if (appConfig.appType === 'development') {
...
+// NOTE: Has to be set up after user data directory is updated
+// https://docs.sentry.io/platforms/javascript/guides/electron/#app-userdata-directory
+Sentry.init({
+	dsn: 'https://f7336c12cc39fb0367886e31036a6cd7@o4507148235702272.ingest.us.sentry.io/4509803831820288',
+	// TODO: Enable tracing based on user consent in production
+	tracesSampleRate: sentryEnvironment === 'production' ? 0 : 1.0,
+	environment: sentryEnvironment,
+	release: appConfig.appVersion,
+	debug: sentryEnvironment === 'development',
+	initialScope: { user: { id: sentryUser.id } },
```
achou11, line 112: would be helpful to have some confirmation that this configured somewhat correctly

src/renderer/src/app.tsx
```diff
@@ -76,6 +82,29 @@ declare module '@tanstack/react-router' {
...
+
+initSentryElectron(
+	{
+		dsn: 'https://f7336c12cc39fb0367886e31036a6cd7@o4507148235702272.ingest.us.sentry.io/4509803831820288',
+		// TODO: Enable tracing based on user consent in production
+		tracesSampleRate: sentryEnvironment === 'production' ? 0 : 1.0,
+		integrations: [tanstackRouterBrowserTracingIntegration(router)],
+		environment: sentryEnvironment,
+		debug: sentryEnvironment === 'development',
+		initialScope: { user: { id: sentryUserId } },
```
achou11, line 103: would be helpful to have some confirmation that this configured somewhat correctly

package.json
```diff
@@ -91,6 +91,8 @@
 		"@ianvs/prettier-plugin-sort-imports": "4.5.1",
 		"@mui/icons-material": "7.2.0",
 		"@mui/material": "7.2.0",
+		"@sentry/react": "9.43",
```
achou11, line 94: there's a more recent version available but this is strictly tied to the version that's needed by `@sentry/electron`. They even provide a useful type error that tells you which SDK you're supposed to install 😄

src/main/index.js
achou11: Thanks, any comments on some of the options that are missing here but are done in the mobile backend's integration? Specifically looking at `enabled`, `sendDefaultPii`, `_experiments`, and the console logging integration.

https://github.com/digidem/comapeo-mobile/blob/fef201da8dcdf93e99c8a567d62ca4834b1a2b9b/src/backend/loader.js#L47-L63

achou11:
Updated this PR to remove the changes to the app related to capturing mutation errors. Those changes now live in #208 as a stacked PR

src/main/index.js
RangerMauve: Regarding enabled it'd be good to have similar semantics to mobile for when it's on/off

I'd set `sendDefaultPii: false` unless you have reason to do otherwise.

Regarding the experiments beforeSendLog, see if the user id is being set in the sentry dashboard with your existing setup first. I had to do that because node.js sentry didn't have a concept of a "user" for logs

src/main/index.js
```diff
@@ -68,14 +73,56 @@ if (appConfig.appType === 'development') {
...
+// We want to make sure that this value is persisted upon startup, even if it might be
+// rotated shortly after.
+configStore.set('sentryUser', sentryUser)
+
+if (shouldRotateSentryUser(sentryUser)) {
+	log('Rotating Sentry user')
+	const newSentryUser = generateSentryUser()
+	configStore.set('sentryUser', newSentryUser)
+	sentryUser = newSentryUser
+}
```
gmaclennan, line 92: Doing this in the `createConfigStore` factory would keep everything in one place and reduce this code in the main index.js

src/main/index.js
```diff
@@ -68,14 +73,56 @@ if (appConfig.appType === 'development') {
...
+} else if (appConfig.appType === 'production') {
+	sentryEnvironment = 'production'
+}
+
+// NOTE: Has to be set up after user data directory is updated
+// https://docs.sentry.io/platforms/javascript/guides/electron/#app-userdata-directory
+Sentry.init({
+	dsn: 'https://f7336c12cc39fb0367886e31036a6cd7@o4507148235702272.ingest.us.sentry.io/4509803831820288',
+	// TODO: Only works on app startup. Any changes to `diagnosticsEnabled` while the app is running will not
+	// take effect here until the app is restarted.
```
gmaclennan, line 108: Nitpick:  ~`TODO`~ -> `NB` - I think it's probably going to be fine to only do this on restart, especially on desktop.

src/preload/main-window.js
```diff
@@ -71,24 +71,37 @@ const runtimeApi = {
...
 }
 
-function getAppVersion() {
-	const flag = process.argv
+/**
+ * @param {`comapeo-${string}`} flag
+ *
+ * @returns {string}
+ */
+function getProcessArgValue(flag) {
```
gmaclennan, line 89: why not use an existing parser like yargs? Or any other argv parser.

src/renderer/src/app.tsx
```diff
@@ -76,6 +82,31 @@ declare module '@tanstack/react-router' {
 	}
 }
 
+let sentryEnvironment: SentryEnvironment = 'development'
+
+if (__APP_TYPE__ === 'release-candidate') {
+	sentryEnvironment = 'qa'
+} else if (__APP_TYPE__ === 'production') {
+	sentryEnvironment = 'production'
+}
```
gmaclennan, line 91: Suggestion: pass this value from the main process, as with the other sentry values, otherwise you risk inadvertently having this as a different value in the main and renderer process, which I think would confuse things.

src/preload/main-window.js
```diff
@@ -71,24 +71,37 @@ const runtimeApi = {
...
 		return ipcRenderer.invoke('settings:set:locale', value)
 	},
+
+	// Sentry
+	getSentryUserId: () => {
+		return getProcessArgValue('comapeo-sentry-user-id')
+	},
+	enableSentryOnInit: () => {
+		return getProcessArgValue('comapeo-sentry-enable') === 'true'
+	},
```
gmaclennan, line 81: suggestions: combine into `getSentryConfig()`, because we are going to be adding more things with the app usage metrics and that will be easier to maintain and keep both processes in sync.

src/renderer/src/app.tsx
```diff
@@ -76,6 +82,31 @@ declare module '@tanstack/react-router' {
...
+const sentryUserId = window.runtime.getSentryUserId()
+const enableSentry = window.runtime.enableSentryOnInit()
+
+initSentryElectron(
+	{
+		dsn: 'https://f7336c12cc39fb0367886e31036a6cd7@o4507148235702272.ingest.us.sentry.io/4509803831820288',
+		enabled: enableSentry,
+		// TODO: Enable tracing based on user consent in production
+		tracesSampleRate: sentryEnvironment === 'production' ? 0 : 1.0,
+		integrations: [tanstackRouterBrowserTracingIntegration(router)],
```
gmaclennan, line 102: I think we should only enable this in production with app usage metrics consent. I know the tracesSampleRate is 0 for production right now, but I think this also adds breadcrumbs to error reports, which some people might consider too much of a privacy issue without consent. That keeps is consistent with mobile.

gmaclennan (commented):
Ok this looks good, the only thing that definitely needs changed is the tanstack routing integration. Other comments are minor suggestions related to maintenance

src/preload/main-window.js
achou11: for a long time, i didn't know that preload scripts had access to installed npm modules. can definitely switch to using a better module for this!

gmaclennan: they can even load native node modules, although the electron docs caution against this

src/main/index.js
achou11: TIL: `NB`. I've been using `NOTE` so will use that for consistency

src/preload/main-window.js
achou11: ah wait nevermind, I misread the documentation [here](https://www.electronjs.org/docs/latest/tutorial/tutorial-preload)

> On top of [Electron modules](https://www.electronjs.org/docs/latest/api/app), you can also access [Node.js built-ins](https://nodejs.org/dist/latest/docs/api/), as well as any packages installed via npm

Thought this section was referring to preload scripts but it's referring to the main process. By default with the sandboxing enabled, preload scripts don't have access to npm-installed modules, hence why I have this custom arg parser.

Really wish they'd expose some of `node:utils` in the preload scripts!

achou11: > they can even load native node modules, although the electron docs caution against this

I think this assumes that the sandboxing is disabled? similar to my realization about normal npm modules. for now I'd rather not disable the sandboxing unless it's blocking essential functionality.

PR #221, opened 2025-08-18
feat: make device discoverable over local network

### LINKED ISSUE #163: Peer discovery (phase 1)
Extracted from https://github.com/digidem/comapeo-desktop/issues/58#issuecomment-2754941953

---

General steps:

- [ ] integrate the [`mdns`](https://github.com/agnat/node_mdns) module into the app and get it to generally run without issues
  - depending on how painful the dev experience is, we may need to consider building out infrastructure for hosting prebuilds
- [ ] set up a service for running mdns and implement something that allows it communicate the necessary information to the core process
  - two potential approaches:
    1. create a mdns service that interfaces with the renderer process. the renderer process then makes the appropriate calls to the core process.
    2. create a mdns service that interfaces with the core process

Goals:

- get something that works on macOS and Linux
- ensure that mobile devices running CoMapeo can be seen and see the desktop device

achou11:
Closes #163 .

~Sets up a new utitility process service that implements advertising itself over the local network to allow others to discover it.~ EDIT: no longer sets up separate service and instead updates the core service.

Does not implement browsing to find other peers yet.

---

Preview:

image


achou11:
moved the discovery stuff into the core process via https://github.com/digidem/comapeo-desktop/pull/221/commits/ef59c5b000bd8540b34f92e1a5e034c3f8d7f912 and also made some general simplifications to the core process implementation via https://github.com/digidem/comapeo-desktop/pull/221/commits/da954b864511834da56384908a8860cb831dd387

PR #257, opened 2025-08-28
chore: rely on using node's built-in runner for running scripts

achou11:
As mentioned in #256 , updating the dev environment to use Node 22.18.0 causes a notable regression in npm script execution time when using npm. To mitigate this issue, I've updated our scripts to be runnable using `node --run`. Note that you can still use `npm run`, but you'll very likely see a notable slowdown depending on which script you're running.

Unfortunately had to replace `npm-run-all2` with `concurrently`, which results in a larger installation (and probably runtime execution) footprint. However, once `npm-run-all2` introduces the ability to use `node --run` internally (see https://github.com/bcomnes/npm-run-all2/issues/155), then we can consider switching back, which I think would be preferable for the most part.


package.json
achou11, line 1: unfortunately there's not really a way to run things sequentially with `concurrently` (makes sense 😄 ). Didn't want to have to install another module so instead opting for `&&`, which might be problematic on Windows? Will try out the builds workflow to test.

achou11: Seems to work based on https://github.com/digidem/comapeo-desktop/pull/257/commits/4c85f68cee3708dfb82bbb25516018c6e6097fbf

PR #259, opened 2025-08-29
chore: use TypeScript for main process code

achou11:
Makes use of the type-stripping functionality that Node backported to 22.18.0 ([details](https://nodejs.org/en/blog/release/v22.18.0)).

Unfortunately, there's a dependency that's pulled in via `electron-store` that doesn't conform to the erasable syntax rules:

```console
types:main     node_modules/json-schema-typed/draft-2020-12.ts(1223,13): error TS1294: This syntax is not allowed when 'erasableSyntaxOnly' is enabled.
```

~I've been wanting to move away from `electron-store` anyways due to it pulling in a ton of deps and functionality that we don't need. Have some local experiments using vanilla zustand as a replacement (weird, i know 😄 ).~ EDIT: Addressed via #260 . The issue mentioned above is no longer relevant!


package.json
```diff
@@ -42,14 +42,15 @@
...
 		"types:services": "tsc -p src/services/tsconfig.json",
 		"types:preload": "tsc -p src/preload/tsconfig.json",
 		"types:renderer": "tsc -p src/renderer/tsconfig.json",
 		"types": "concurrently --prefix \"types:{name}\" --pad-prefix --group \"node:types:*\"",
-		"test": "concurrently --pad-prefix --group \"node:types\" \"node:vitest:run\"",
-		"vitest:run": "vitest run --config ./src/renderer/vite.config.js",
-		"vitest:watch": "vitest watch --config ./src/renderer/vite.config.js",
+		"test:renderer": "vitest run --config ./src/renderer/vite.config.js",
+		"test:shared": "vitest run --dir ./src/shared/",
+		"test": "concurrently --prefix \"test:{name}\" --pad-prefix --group \"node:test:*\"",
```
achou11, line 53: Eventually should be using [Vitest projects](https://vitest.dev/guide/projects.html) as opposed to separate commands (I think). Something to address later on!

PR #267, opened 2025-09-03
chore: improve releases workflow

### LINKED ISSUE #192: Upload release assets in `release` workflow instead of `create-builds` workflow
Extracted from https://github.com/digidem/comapeo-desktop/pull/190#discussion_r2258230179

Currently, the uploading of builds to a release is conditionally done in the `create-builds` workflow. Ideally this should happen in the `release` workflow, after all of the builds have completed.

The main trickiness is making sure all of the correct artifacts are uploaded.

My initial thinking on an approach is roughly:

1. use `actions/download-artifact` to download the relevant artifacts (use their merge option)
2. upload the artifacts using either:
  - a custom forge publisher plugin
  - the github cli

achou11:
Closes #192 

- Updates to `create-builds` workflow:
  - removes release upload step

- Updates to `release` workflow:
  - adds step to download and merge artifacts from the `create-builds` workflow
  - adds step to upload assets to relevant github release

- Removes any code and docs related to publishing releases using Forge, as everything is now handled via workflows using the github cli.


achou11:
Based on result of https://github.com/digidem/comapeo-desktop/actions/runs/17440675506, still need to adjust the artifact merging. The per-OS upload preserves the directory structure since we're using the globbing option. Need to figure out how to flatten it 😅 

image

.github/workflows/create-builds.yml
```diff
@@ -98,21 +100,23 @@ jobs:
...
+    outputs:
+      artifact_id: ${{ steps.merge_and_upload.outputs.artifact-id }}
+    steps:
+      - name: Merge and upload artifacts
+        id: merge_and_upload
+        uses: actions/upload-artifact/merge@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: builds-${{ env.BUILD_SHA }}
+          delete-merged: true
+          compression-level: 0
```
achou11, line 122: wondering if it would be preferable to do this in the release workflow instead of here. Being able to have matrix-separated artifacts for the create-builds workflow is kind of helpful for debugging and reusability.

Reason I didn't do it is is that it's harder to test that this generally works, as it would require me to run a "test" release workflow.

forge.config.ts
```diff
@@ -357,6 +348,30 @@ export default {
 
 			return packageJson
 		},
+		// NOTE: Creates an output directory at `out/artifacts` that contains just the distributable assets.
+		// Used by the [`create-builds`](./github/workflows/create-builds.yml) workflow
+		postMake: async (config, makeResults) => {
```
achou11, line 353: is it worth doing this only when `process.env.CI` is set?

PR #272, opened 2025-09-04
chore(deps): upgrade deps

achou11:
Held off on upgrading [Electron to v38](https://releases.electronjs.org/release/v38.0.0), at least until better-sqlite3 publishes a release with compatible prebuilds.


achou11:
For posterity: created an internal build via https://github.com/digidem/comapeo-desktop/actions/runs/17469462467 and ran it. No new notable issues

PR #280, opened 2025-09-09
chore: update windows packaging configuration

achou11:
Came across this detail in the [Forge docs](https://www.electronforge.io/config/makers/squirrel.windows#spaces-in-the-app-name) and it became a small rabbithole of updating the app configuration to account for it.

> Squirrel.Windows can behave unexpectedly when application names contain spaces.

There's a change to the app config schema as a result but didn't treat it as something worth writing a migration for, given that it's optional, only applies to windows, and we're still in the pre-release phase.

Still need to test that these changes work on Windows (currently don't have my Windows laptop with me).

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -433,6 +433,7 @@ function getPropertiesForAppType(appType: AppType): {
 			const appVersion = getAppVersion({
 				version: packageJSON.version,
 				appType,
+				appTypeSuffix,
 			})
 
 			return {


achou11:
For posterity, was able to test these changes on my Windows machine. Overall pretty satisfied with the results. The Forge docs I pointed to are a bit inaccurate, causing me to spend way more time that I hoped on getting the app user model ID stuff right 😅 

Also seemed to have discovered a bug (may be something related to Electron). Will create an issue shortly

PR #293, opened 2025-09-16
fix: fix application not fully closing on Windows and Linux

### LINKED ISSUE #283: App process lingers when quitting app on Windows
Reproduction steps:

1. Open the task manager (`ctrl + shift + esc` if you prefer keyboard shortcut)

2. Use `comapeo` in text input to filter by app name

3. Start CoMapeo application. You should see the app under the `Apps` section in the listed processes

4. Do either of the following to close the application:

  a. Use the keyboard shortcut (`alt + f4`)
  b. Use the close button in the top right of the window

5. In the task manager, notice that the application process has moved to the `Backgrounded processes` section. 

From my understanding, being a background process means that the app is still technically running. I came across issues when attempting to delete relevant app files via the terminal because the app is technically still using them.

What's interesting is that if you close the application using the `File > Exit` menu item, then this behavior in (5) doesn't occur! The process completely disappears as expected. That menu item basically calls `app.quit()` directly, and it makes me wonder if misbehaving options in (4) do something different under the hood 🤔 I want to believe that this could be an Electron issue, but wouldn't be surprised if there's some problematic code in our core service (or main process).

achou11:
Fixes #283

### DIFF
--- a/src/main/app.ts
+++ b/src/main/app.ts
@@ -248,7 +248,7 @@ function initMainWindow({
 	}
 
 	mainWindow.on('close', (event) => {
-		if (!APP_STATE.tryingToQuitApp) {
+		if (process.platform === 'darwin' && !APP_STATE.tryingToQuitApp) {
 			log(`Hiding main window with id ${mainWindow.id}`)
 			event.preventDefault()
 			mainWindow.hide()


achou11:
Confirmed to work on Linux and Windows in development, and with internal builds from https://github.com/digidem/comapeo-desktop/actions/runs/17803613304

PR #350, opened 2025-10-22
chore: enable Sentry in core service

achou11:
Had no idea that I needed to explicitly set up Sentry in utility processes as I thought that was handled via the primary electron SDK 😅 In all fairness, it's not really documented and I had to figure this out by looking at https://github.com/getsentry/sentry-electron/pull/991.


achou11:
Created an issue in Sentry repo: https://github.com/getsentry/sentry-electron/issues/1254

PR #493, opened 2026-01-21
chore: implement e2e test isolation

achou11:
Lays the groundwork for being able to run e2e tests in isolation from one another, which enables the following:

1. Being able to run the tests in parallel. Before, there was a specific sequence that needed to occur between tests because of the use of outputs between test projects. Now each test runs with its own data directory so there's no sharing of outputs.

2. Being able to "parameterize" tests such that we can test various scenarios for a given flow. The old test setup was pretty limited in that we were essentially testing a very specific flow (e.g. single player project creator) and if we wanted to test a different flow, it would require a pretty hard to manage implementation. The isolation allows us to write tests that set up their own unique environment/situation.

Notable implementation details:

- All e2e tests are run in parallel. When running locally, we allow Playwright to attempt to use as many CPU cores as are available. In CI we limit it to 3 max, which is based on the [specs of the GitHub Actions runners](https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/choose-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories).

- Each test runs an instance of the application that uses its own user data directory, which provides data isolation.


achou11:
Seeing this warning on windows runs:

```console
(node:1492) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
```

[Issue upstream](https://github.com/microsoft/playwright/issues/38278) was addressed but hasn't been released, it seems.

PR #571, opened 2026-04-09
chore: update @comapeo/core to v7

achou11:
Updates comapeo deps that align with @comapeo/core@7. There are no notable userfacing changes resulting from the upgrade, but it required many internal changes like:

- usage of `@comapeo/map-server` for handling maps instead of the maps implementation in core
- accounting for change in default return value when retrieving many members of a project
- setup of app rpc
- removal of unnecessary hooks due to updates in map setup

Also adds a few patches related to issues/limitations I ran into, mostly related to `@comapeo/map-server`. Relevant upstream issues:

- https://github.com/digidem/comapeo-map-server/issues/43
- https://github.com/digidem/comapeo-map-server/issues/45


src/renderer/src/hooks/maps.ts
```diff
@@ -1,36 +1,51 @@
...
-				update('maps')
+		function updateRefreshTokenWhenOnline() {
+			let timeoutId: number | undefined
+
+			function handleOnOnline() {
+				// TODO: Should eventually remove this delay
+				// https://github.com/digidem/comapeo-map-server/issues/44
+				timeoutId = window.setTimeout(() => {
+					setRefreshToken(Date.now().toString())
+				}, 3_000)
```
achou11, line 21: ideally unnecessary to do this. tracking upstream in https://github.com/digidem/comapeo-map-server/issues/44

src/renderer/src/rpc-clients.ts
```diff
@@ -0,0 +1,29 @@
+import { createAppRpcClient, createMapeoClient } from '@comapeo/ipc/client.js'
+
+export function initRpcClients() {
+	const comapeoChannel = new MessageChannel()
+	const appChannel = new MessageChannel()
```
achou11, line 5: might be overkill to set up a separate message channel for the app rpc since ipc sets up subchannels anyways. the limitation around using a single channel is less relevant for desktop than mobile so think it's okay to have a separate one in this case.

PR #581, opened 2026-04-17
chore: implement major update to translations setup

### LINKED ISSUE #499: Overhaul translations setup
# Details

There are a variety of limitations with how our translations are set up at the moment. Gregor and I started experimenting with a few improvements to our process as an org and how it maps to the technical pieces of integrating translations. Some takeaways (so far):

## Consolidation of the desktop and mobile projects on Crowdin

Mostly for the sake of Crowdin project management. The only anticipated technical consequences are: 

- updates to the repo's crowdin config
- updates to the github integration
- updates to the repo's message extraction tooling

## We need to introduce some kind of category indicating importance of a given message

The early idea was to have 3 categories: core, primary, and secondary. These indicate a message's importance to the usage of the app and can be used as a concrete measurement of how "ready" a language is to be used within the app. For example, we could define a usable language as a language that has _all core_ messages translated, meaning that it will be surfaced as an option for the language settings. Having the additional categories of primary and secondary is mostly useful for translators to understand prioritization.

We can potentially use this [feature in formatjs](https://formatjs.github.io/docs/tooling/babel-plugin#pragma) in order to facilitate categorization of messages and how they're extracted for usage in Crowdin.

# Designs

N/A

achou11:
Closes #499

Notable changes:

- Migrates existing translations to have a folder structure such that there's a directory for each language, where each contains a `primary.json` and a `secondary.json` file.
  - Also fixes up existing non-English translations such that they no longer include English placeholders.
- Updates app code to use "special" message IDs where (potentially appropriate).
  - I opted for categorizing the majority of messages as "primary" - basically anything that visibly surfaces in the app. This will most likely change based on a more proper audit later.
- Updates the extraction and compiling tooling to account for special message IDs and updated translations folder structure.
- Updates app code to account for the usage of the regional variants of languages.
  - For now, language settings do not surface the regional specificity of the languages. This will be considered when we have actual support for multiple regional variants for the same language.
  - For existing users, we are only storing the base language tag. In general, we want to be persisting the regional variant of the selected language moving forward. This PR implements a migration step in the persisted store that will migrate the persisted base language tag to its regional counterpart if applicable. See changes in persisted store implementation and related tests.


src/main/persisted-store.ts
```diff
@@ -166,6 +166,24 @@ export function createPersistedStore(opts: { filePath: string }) {
 		store.setState({ appUsageMetrics: updatedAppUsageMetrics })
 	}
 
+	// NOTE: We started to include the region tags for persisted languages after initial release
```
achou11, line 169: highlighting persisted locale migration

src/renderer/src/lib/intl.ts
```diff
@@ -3,33 +3,70 @@ import * as v from 'valibot'
...
 ): Promise<NonNullable<IntlConfig['messages']>> {
-	// @ts-expect-error Not worth fixing
-	return translations[`./${language}.json`]!()
+	const [baseTag, regionTag] = languageTag.split('-')
+
+	if (!baseTag) {
+		throw new Error(`Cannot get base tag from ${languageTag}`)
+	}
+
+	const relevantTranslations = Array.from(Object.entries(translations))
```
achou11, line 44: noting that when a language is selected, we load the translations for anything sharing the base language, with the regional variant of the selected language taking highest precedence. For example, if we have something like `es`, `es-ES`, and `es-419`, and `es-ES` is selected, then the contruction of the translations object will be `{ ...es, ...es-419, ...es-ES }`.

this is to provide a better fallback behavior than just resorting to the default English translations in the case of missing keys.

src/renderer/src/routes/app/settings/_nested/language.tsx
```diff
@@ -94,15 +92,31 @@ function RouteComponent() {
...
+										englishName,
+										nativeName,
+										baseLanguageInfo,
+									}) => (
 										<FormControlLabel
 											key={languageTag}
 											value={languageTag}
 											control={<Radio />}
 											label={
+												// NOTE: We intentionally do not show the regional variant for now.
```
achou11, line 106: Noting this is where we surface just the base language name in the language settings i.e. without the regional variant, which is subject to change in the future.

src/renderer/src/routes/app/settings/index.tsx
```diff
@@ -112,7 +113,12 @@ function SettingsList() {
 	const { data: selectedLanguageName } = useSuspenseQuery({
 		...getLocaleStateQueryOptions(),
 		select: ({ source, value }) => {
-			const match = getLanguageInfo(value)
+			const baseTag = value.split('-')[0]!
+
+			// NOTE: We intentionally do not show the regional variant for now.
```
achou11, line 118: Same note about only showing the base language name and not the regional component, for now.

PR #594, opened 2026-05-28
chore: compile translations as a separate workflow

### LINKED ISSUE #592: Consider dropping support for macOS x64
Ideally doing this will be determined by limitations around the availability of github runners, but the surrounding ecosystem may also be a factor (e.g. https://github.com/formatjs/formatjs/issues/6602). It may come to a point where we need to consider dropping support for it for the sake of making maintenance more straightforward. The runner has also been known to be slower and a little less reliable, so removing that from our matrix would be a plus there.

In the case of the formatjs situation, the alternative approach is to move away from it to something like https://lingui.dev/, which is something I've been considering doing anyways. It's a much bigger lift though.

achou11:
Closes #592

Creates a reusable workflow for compiling translations, which is then used by dependent jobs. This avoids platform limitations of the tooling, as the output of the compilation is platform-agnostic.

Also does a lot of cleanup related to improving security for affected workflows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,9 +66,6 @@ jobs:
           artifact-ids: ${{ needs.compile_translations.outputs.artifact_id }}
           path: translations/
 
-      - name: Generate translated languages manifest
-        run: node --run intl:generate
-
       - name: Run checks
         run: node --run checks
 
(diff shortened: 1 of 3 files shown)


achou11:
Confirmed that `create-builds` workflow and its outputs work as expected via https://github.com/digidem/comapeo-desktop/actions/runs/26588417627

PR #634, opened 2026-07-22
feat: expose observation metadata

### LINKED ISSUE #624: Expose observation metadata
# Details

Feature parity with mobile, exposing location-related data tied to an observation.

# Design

Image

Image

achou11:
Closes #624 

- The rounding of values displayed is based on comments from Gregor ([Slack link](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1784733235260269?thread_ts=1784666507.807119&cid=C6ZMBFATZ)):

    > I think lat and lon should be rounded. There is no such thing as a "true value" and there is nothing gained by showing more. The 7th decimal place is 1.1cm on the equator. The 4th decimal place is 11m, which is about the limit of the phone's accuracy, but I think 5 decimal places is good - 1.1m (GPS will never be as accurate as that) and it aligns with the accuracy of UTM coordinates rounded to the nearest decimal place (UTM coordinates are meters, more or less).
    > 
    > Speed should be rounded to 1 decimal place, although I think it would be better to convert it to kmph or mph (depending on locale settings) and round it to 1 decimal place < 10mph and no decimals beyond that.
    > 
    > Accuracy should be rounded to the nearest integer - accuracy is not very accurate from a GPS!
    > 
    > Altitude to the nearest meter / foot.
    
    I haven't implemented the speed suggestions because it probably differentiates from mobile a bit too much. Can do as a dedicated follow-up. For now, I've rounded it to 2 decimal places as mobile currently does.

---

image

image


src/renderer/src/components/app-title-bar.tsx
achou11, line 1: for my future reference, changes here ensure that tooltips never overlap app title bar

Image

PR #648, opened 2026-08-04
fix: fix issue with undefined project API in some project-specific loaders

achou11:
Fixes an issue where we cannot access the project API instance in a route `onEnter`/`onLeave` loader hook in some project-specific routes.

I wasn't seeing this on my local macOS machine but was able to on my Linux machine. Not sure how this hasn't been experienced before, but here's a recent [Sentry event](https://awana-digital.sentry.io/issues/7583195059/events/17b5238ab00b4e9a93a4075f832e0da4/) seen during QA.

Seems to be an issue in Tanstack Router (see https://github.com/TanStack/router/issues/3293) that's fixed in a [prerelease](https://github.com/TanStack/router/releases/tag/release-2026-07-24-1714) version.


achou11:
Ah looks like upstream created a stable release with the fix: https://github.com/TanStack/router/releases/tag/release-2026-08-04-2036

Nice timing 😄

PR #664, opened 2026-08-26
fix: expose all supported languages as options in language settings

### LINKED ISSUE #597: Expose language options that we claim to support even if we don't have translations for them
Came across this [comment in mobile](https://github.com/digidem/comapeo-mobile/blob/ddc0deec5731290c6b2b1c52dcabbdc0003ba206/src/frontend/lib/intl.ts#L41-L49) that describes a very relevant problem: the translation of a categories set is completely independent of the translation of the app. As of right now, if someone uses a categories set that's translated in some language _L_ but there are no _app_ translations for that language, then there's no option for _L_ to be used in the app (see [relevant script](https://github.com/digidem/comapeo-desktop/blob/8cd915b3ef47d5b3e8710fff6b1659fe7e42850b/scripts/generate-translated-languages-list.ts#L60-L63) that generates the "translated" languages list used to determine the app language setting options). This means that the user would not be able to make use of the translations used by the categories set.

The initial reasoning for not exposing languages with zero app translations was because of feedback about how confusing the UX was when selecting said language and there being no visible changes in the app (believe this feedback was in the context of mobile but applies to desktop too). Guess it becomes a lesser of two evils choice, where my guess is that that not being able to access categories set translations is worse than said confusing UX.

The simple initial solution is to probably remove the check from the referenced script that filters out translations-less languages. To mitigate the UX consequence, we probably have to incorporate smarter inclusion of fallback languages based on things like the system preferences, browser language, etc (separate but related limitation that probably deserves its own issue).

I also wonder if there's a way to get a list of languages that a categories set is translated in. That could be a useful bit of information that helps with exposing language options in the app settings independently of the app translations status. For example, if we had this kind of information, there could be a UX where we expose a separate setting related to the language to use for collected data, where options would be either to follow the _app language_ or the list of languages that the categories set is translated in.

achou11:
Closes #597 

We now surface all languages that we attempt to provide translations. Effectively, this is determined by our configuration in Crowdin and whatever languages get inserted into the `messages/` directory for the repo (regardless of the existence of actual translations).


src/renderer/src/contexts/intl.tsx
```diff
@@ -62,22 +67,40 @@ export function IntlProvider({ children }: PropsWithChildren) {
...
+	// 1. Locale resolved by the backend
+	// 2. What the browser detects using the built-in web APIs
+	// 3. App fallback (en-US)
+	//
+	// The value returned by (1) is calculated using a similar approach but using Electron's APIs
+	// for getting the system-preferred languages (see main/intl-manager.ts).
+	const languagesToUse: Array<SupportedLanguageTag> = Array.from(
+		new Set([
+			deferredLocale,
+			...navigator.languages.filter((l) => is(SupportedLanguageTagSchema, l)),
```
achou11, line 85: kind of assumes that this returns something similar to what Electron's API for getting system-preferred languages does. reasons for not forwarding from main process to renderer instead are:

1. Less lock-in for the renderer app. if the app were ever truly an isolated web-app, it ideally uses the native browser apis for this, as is being done here. Sticking to principles based on lofty scenarios 😄 

2. Less work on the IPC layer.

Likelihood of drift between the two seems unlikely and not catastrophic, so willing to live with the risk for now until proven necessary.

messages/gu/primary.json
achou11, line 1: `gu/` directory was leftover from the translation tooling migration work. Our Crowdin is set up to only update the `gu-IN/` directory now.

